PR #2, opened 2019-02-12
Subclass Group and use new accessors.

### LINKED ISSUE #1: allow changing text with setter
dynamic changin' of text would be rad

tannewt:
Fixes #1 

This depends on https://github.com/adafruit/circuitpython/pull/1535


ladyada:
tested - so far so good!

PR #5, opened 2019-03-02
Add line_spacing property and change (0,0) definition

ladyada:
the origin point of text areas is from the baseline of the font rather than the font's bounding box corner - that way changing fonts will not change the location of the text, only size and style

see this for definition of font baseline
https://printingpartners.files.wordpress.com/2017/12/font-diagram.png

see pygame which does the same (i think) https://nerdparadise.com/programming/pygame/part5

@tannewt its a breaking change so i'd like to do this soon if you're OK with it


ladyada:
@tannewt this is the new version with center-M origin

adafruit_display_text/label.py
```diff
@@ -44,21 +44,20 @@
 __version__ = "0.0.0-auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
 
-class TextArea(displayio.Group):
+class Label(displayio.Group):
     """An area displaying a string of textself.
```
tannewt: Please add a comment about how the text is positioned relative to the position.

ladyada:
@tannewt i need to lint more

adafruit_display_text/label.py
```diff
@@ -44,21 +44,23 @@
 __version__ = "0.0.0-auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
 
-class TextArea(displayio.Group):
-    """An area displaying a string of textself.
+class Label(displayio.Group):
+    """An area displaying a string of textself. The origin point set by ``x`` and ``y``
```
tannewt:
```suggestion
    """A label displaying a string of text. The origin point set by ``x`` and ``y``
```

Issue #15, opened 2019-07-21
display_text_simpletest.py is blank, not functional.


jtrip:
Hello, while looking into this library the first time I noticed that the display_text_simpletest.py is blank. The README.md file illustrates a simple test that I think would be great, however I would also like to suggest making one small improvement.

Currently the usage example is:  
```
import board
import terminalio
from adafruit_display_text import label

text = "Hello world"
text_area = label.Label(terminalio.FONT, text=text)
text_area.x = 10
text_area.y = 10
board.DISPLAY.show(text_area)
```

Which does seem to work in my testing, however after updating the display with the text_area object the code immediately ends which leads to circuitpython displaying the message that the code is done running. My only concern is that beginners might not have the intuition to think that it is working fine but some additional control needs to be used in order for it to be visible for more than a moment, and then the beginner might be chasing down a ghost thinking it isn't working right.  

I think a beginner friendly improvement would be to add a delay so that they should be able to see the output but then also see that when the code is done executing it will be replaced by the circuit python "Code done running" message.   
Like so:  
```
import board
import terminalio
from adafruit_display_text import label
from time import sleep

text = "Hello world"
text_area = label.Label(terminalio.FONT, text=text)
text_area.x = 10
text_area.y = 10
board.DISPLAY.show(text_area)
sleep(3)
```
Another simple option might be to put the board.DISPLAY.show call in a loop (maybe infinite like `while True:`) but maybe that's not setting a great example for beginners or is otherwise adding irrelevant logic to the example.

I would be happy to make these changes and submit a pull request. I don't think I have submitted a pull request to any of your repos yet so I thought I would introduce the idea here. Perhaps there is some reason that example is currently missing or if you have any thoughts about the change please let me know.

ladyada:
oops looks like we missed that example, please feel free to submit a nice simple one. what youv've got looks great. instead of a sleep, please do a 'while True: pass' type loop so it stays on screen (or you can bounce the text around :D

jtrip:
With that merged in I think we are good here, thanks!

PR #16, opened 2019-07-22
Update simple example and usage example in readme

jtrip:
Update simple example and usage example in readme, additional line after imports for pep 8.

As mentioned in #15 the simplest test example was blank although there was a good example in the readme file. I update the simple test file to match the usage example in the readme file and also put the display updating call in a `while True:` loop so a user will be sure to see the output.

I also added an additional blank line after the import statements in that file for pep 8 compliance. I also then copied that code to the usage example in the readme so that the update is reflected there.

I built the docs to verify the changes were reflected there, they were and I saw no other relevant aspect of the docs that needed updating. Even though the simple test file was blank the docs were already pointing to it, probably because there are so many of these wonderful libraries. 😄 

If the pep 8 line is not welcome I can remove that and resubmit.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,11 +36,13 @@ For a board with a built-in display.
     import terminalio
     from adafruit_display_text import label
 
+
     text = "Hello world"
     text_area = label.Label(terminalio.FONT, text=text)
     text_area.x = 10
     text_area.y = 10
-    board.DISPLAY.show(text_area)
+    while True:
+        board.DISPLAY.show(text_area)
 
 
 Contributing
--- a/README.rst
+++ b/README.rst
@@ -41,8 +41,9 @@ For a board with a built-in display.
     text_area = label.Label(terminalio.FONT, text=text)
     text_area.x = 10
     text_area.y = 10
+    board.DISPLAY.show(text_area)
     while True:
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -36,11 +36,13 @@ For a board with a built-in display.
     import terminalio
     from adafruit_display_text import label
 
+
     text = "Hello world"
     text_area = label.Label(terminalio.FONT, text=text)
     text_area.x = 10
     text_area.y = 10
-    board.DISPLAY.show(text_area)
+    while True:
```
ladyada: show and then while True: pass

examples/display_text_simpletest.py
```diff
@@ -0,0 +1,11 @@
+import board
+import terminalio
+from adafruit_display_text import label
+
+
+text = "Hello world"
+text_area = label.Label(terminalio.FONT, text=text)
+text_area.x = 10
+text_area.y = 10
+while True:
```
ladyada: ditto

jtrip:
Ah yes, sorry I misunderstood you the first time you described it. Having it do less makes sense to me!  
That should be good now.

ladyada:
@makermelissa please review when you can

PR #18, opened 2019-08-19
Updated PyPortal example to work with latest stable CircuitPython

### LINKED ISSUE #17: display_text_pyportal.py outdated
It uses the old style for the backlight pin and is from a demo rev.

makermelissa:
Fixes #17.

### DIFF
--- a/examples/display_text_pyportal.py
+++ b/examples/display_text_pyportal.py
@@ -48,5 +48,4 @@ for b in range(100, -1, -1):
 
 print("fade down")
 
-while True:
-    pass
+time.sleep(10)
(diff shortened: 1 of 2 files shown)


examples/display_text_pyportal.py
```diff
@@ -40,14 +38,15 @@
...
-    time.sleep(0.005)  # default (0.005)
+for b in range(100, -1, -1):
+    board.DISPLAY.brightness = b / 100
+    time.sleep(0.01)  # default (0.01)
 
 print("fade down")
 
-    # splash.pop()
+while True:
+    pass
```
tannewt: Why wait if the code is done? Isn't it ok to kick to the terminal?

makermelissa: Either way. It just shows the fade out nicely, but perhaps another delay and then kick back to terminal?

tannewt: Ya, sounds good.

PR #30, opened 2020-01-29
adding anchored position example

FoamyGuy:


examples/display_text_anchored_position.py
```diff
@@ -0,0 +1,67 @@
...
+text_group.append(text_area_top_right)
+text_group.append(text_area_middle_middle)
+text_group.append(text_area_middle_left)
+text_group.append(text_area_middle_right)
+text_group.append(text_area_bottom_middle)
+text_group.append(text_area_bottom_left)
+text_group.append(text_area_bottom_right)
+
+print(text_area_top_right.anchored_position)
+print(text_area_top_right.anchor_point)
```
caternuson: still need these prints?

examples/display_text_anchored_position.py
```diff
@@ -0,0 +1,67 @@
...
+
+text_area_middle_middle = label.Label(terminalio.FONT, text=text)
+text_area_middle_middle.anchor_point = (0.5, 0.5)
+text_area_middle_middle.anchored_position = (DISPLAY_WIDTH/2,DISPLAY_HEIGHT/2)
+
+text_area_middle_right = label.Label(terminalio.FONT, text=text)
+text_area_middle_right.anchor_point = (1.0, 0.5)
+text_area_middle_right.anchored_position = (DISPLAY_WIDTH-10, DISPLAY_HEIGHT/2)
+
+
```
caternuson: extra line

examples/display_text_anchored_position.py
```diff
@@ -0,0 +1,67 @@
...
+import board
+import terminalio
+from adafruit_display_text import label
+import displayio
+
+DISPLAY_WIDTH = 320
+DISPLAY_HEIGHT = 240
+
+text_group = displayio.Group(max_size=9)
+text = "Hello"
```
caternuson: This is like a const, so maybe change it to `TEXT` and move it up by the `DISPLAY_` defs?

examples/display_text_anchored_position.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+import board
+import terminalio
+from adafruit_display_text import label
+import displayio
+
+DISPLAY_WIDTH = 320
+DISPLAY_HEIGHT = 240
+
+text_group = displayio.Group(max_size=9)
```
caternuson: This might be better moved to just before your first `text_group.append` line below.

examples/display_text_anchored_position.py
FoamyGuy: I would say the prints are not really essential, I mainly put them to test and ensure the getters were working as expected. I can remove them.

examples/display_text_anchored_position.py
FoamyGuy: Yep, agreed. Will do.

examples/display_text_anchored_position.py
FoamyGuy: I like it, better to keep the `text_group` setup together than have it split by all of the lines setting up labels. I will make this change along with the others in the next commit

caternuson:
Thanks for the updates. Looks good.

PR #31, opened 2020-02-01
Convert text object to string.

kattni:
Allows for providing float object, such as acceleration or gyro data, without needing to convert to string in user code.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -80,7 +80,7 @@ class Label(displayio.Group):
         self._boundingbox = None
 
         if text:
-            self._update_text(text)
+            self._update_text(str(text))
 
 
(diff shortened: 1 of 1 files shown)


jepler:
In chat with kattni, I mentioned that maybe this would affect how bytes objects were displayed.  I was wrong about that (as of now, it's an error to pass a bytes object at all), but it *does* affect how lists and tuples of strings would be displayed.

Before this PR, `l.text = ['a', 'b', 'c']` would behave approximately as `l.text = 'abc'`, but after it would show as the bracketed list.  Similar for tuples.

I don't think this necessarily means this PR shouldn't be merged, because to me it is very defensible that you could use "anything print()able" with a Label, even if the property is called "text".

ladyada:
yah folks should format() to hearts desire if they dont want str()

Issue #33, opened 2020-02-08
Label could have a background color


dglaude:
Right now only the text color can be selected, and the background color is transparent.
Being able to optionally define the background color would be great.

PR #35, opened 2020-02-15
Adding background_color

FoamyGuy:
This change allows you to set background_color of the label in the constructor or with a property. Setting background_color to false, or excluding it entirely will result in same current behavior which is transparent background. This is referenced in issue #33.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -72,7 +72,7 @@ class Label(displayio.Group):
 
         self.palette = displayio.Palette(2)
         if background_color or background_color == 0x000000:
-            self.palette[0] = new_color
+            self.palette[0] = background_color
             self.palette.make_opaque(0)
             self._transparent_background = False
         else:
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -44,7 +44,6 @@ import displayio
 __version__ = "0.0.0-auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
 
-
 class Label(displayio.Group):
     """A label displaying a string of text. The origin point set by ``x`` and ``y``
        properties will be the left edge of the bounding box, and in the center of a M
@@ -57,7 +56,6 @@ class Label(displayio.Group):
        :param int max_glyphs: The largest quantity of glyphs we will display
        :param int color: Color of all text in RGB hex
        :param double line_spacing: Line spacing of text to display"""
-
     def __init__(self, font, *, x=0, y=0, text=None, max_glyphs=None, color=0xffffff,
                  backgroud_color=False, line_spacing=1.25, **kwargs):
         if not max_glyphs and not text:
@@ -87,14 +85,15 @@ class Label(displayio.Group):
         if text is not None:
             self._update_text(str(text))
 
-    def _update_text(self, new_text):  # pylint: disable=too-many-locals
+
+    def _update_text(self, new_text): # pylint: disable=too-many-locals
         x = 0
         y = 0
         i = 0
         old_c = 0
         y_offset = int((self.font.get_glyph(ord('M')).height -
                         new_text.count('\n') * self.height * self.line_spacing) / 2)
-        # print("y offset from baseline", y_offset)
+        #print("y offset from baseline", y_offset)
         left = right = top = bottom = 0
         for character in new_text:
             if character == '\n':
@@ -104,10 +103,10 @@ class Label(displayio.Group):
             glyph = self.font.get_glyph(ord(character))
             if not glyph:
                 continue
-            right = max(right, x + glyph.width)
-            if y == 0:  # first line, find the Ascender height
-                top = min(top, -glyph.height + y_offset)
-            bottom = max(bottom, y - glyph.dy + y_offset)
+            right = max(right, x+glyph.width)
+            if y == 0:   # first line, find the Ascender height
+                top = min(top, -glyph.height+y_offset)
+            bottom = max(bottom, y-glyph.dy+y_offset)
             position_y = y - glyph.height - glyph.dy + y_offset
             position_x = x + glyph.dx
             if not self._text or old_c >= len(self._text) or character != self._text[old_c]:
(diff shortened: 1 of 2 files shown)


adafruit_display_text/label.py
```diff
@@ -57,7 +57,7 @@ class Label(displayio.Group):
        :param int color: Color of all text in RGB hex
        :param double line_spacing: Line spacing of text to display"""
     def __init__(self, font, *, x=0, y=0, text=None, max_glyphs=None, color=0xffffff,
-                 line_spacing=1.25, **kwargs):
+                 backgroud_color=False, line_spacing=1.25, **kwargs):
```
makermelissa: I think background should have an extra 'n' in it ;)

FoamyGuy: whoops, thank you. Correction pushed now.

makermelissa:
Hi, looks mostly good. However, if you don't set the background color so that it is transparent and read the property, it returns 0, which is the same value returned if it is black. Can we modify this so that if the color is transparent, it returns None?

It doesn't appear circuitpython currently has a native way to check whether it is transparent, so we may need to store that as a flag on the label internally.

adafruit_display_text/label.py
```diff
@@ -57,7 +57,7 @@ class Label(displayio.Group):
        :param int color: Color of all text in RGB hex
        :param double line_spacing: Line spacing of text to display"""
     def __init__(self, font, *, x=0, y=0, text=None, max_glyphs=None, color=0xffffff,
-                 line_spacing=1.25, **kwargs):
+                 background_color=False, line_spacing=1.25, **kwargs):
```
caternuson: Can you change the default to `None` instead and then change your logic to match? You can check for `None` easily:
```python
if background_color:
    # action if there is a color set
else:
    # action if it's None
```

examples/display_text_background_color.py
```diff
@@ -0,0 +1,15 @@
...
+import board
+import terminalio
+from adafruit_display_text import label
+
+
+text = " Color Background Hello world"
+text_area = label.Label(terminalio.FONT, text=text, color=0x0000FF, background_color=0xFFAA00)
+text_area.x = 10
+text_area.y = 10
+board.DISPLAY.show(text_area)
```
caternuson: Also add some lines that show:
* querying current background color
* changing back to no background color

adafruit_display_text/label.py
```diff
@@ -168,6 +171,19 @@ def color(self):
     def color(self, new_color):
         self.palette[1] = new_color
 
+    @property
+    def background_color(self):
+        """Color of the background as an RGB hex number."""
+        return self.palette[0]
```
caternuson: It would be good if this would return `None` when there is no background color set.

FoamyGuy:
@makermelissa @caternuson I believe all of the mentioned changes have been made with the new commits. In making and testing the changes I also found and fixed a few issues with changing too and from transparent background colors.

makermelissa:
Thanks @FoamyGuy. One more small thing. Can we change it so `self.transparent_background` is private so that it becomes `self._transparent_background`. That way you don't have users doing something silly like:
```
text_area = label.Label(terminalio.FONT, text="hello", color=0x0000FF)
text_area.transparent_background = False
print(text_area.background_color)
```

FoamyGuy:
@makermelissa Got it done. Thank you.

adafruit_display_text/label.py
```diff
@@ -71,7 +71,12 @@ def __init__(self, font, *, x=0, y=0, text=None, max_glyphs=None, color=0xffffff
         self.y = y
 
         self.palette = displayio.Palette(2)
-        self.palette.make_transparent(0)
+        if not background_color:
+            self.transparent_background = True
+            self.palette.make_transparent(0)
+        else:
+            self.transparent_background = False
+            self.palette[0] = background_color
```
caternuson: Sorry, I totally forgot about needing to accept `0` as a possible valid color. If that were done with above, then it would be treated as setting transparent. How about this?
```python
        if background_color is not None:
            self.palette[0] = background_color
            self.palette.make_opaque(0)
            self.transparent_background = False
        else:
            self.palette.make_transparent(0)
            self.transparent_background = True
```

Or could maybe do like you did in the setter?

caternuson: (but also with the leading underscore)

FoamyGuy: Ah right, thank you. just pushed some commits to fix this in the constructor.

adafruit_display_text/label.py
```diff
@@ -71,7 +71,14 @@ def __init__(self, font, *, x=0, y=0, text=None, max_glyphs=None, color=0xffffff
         self.y = y
 
         self.palette = displayio.Palette(2)
-        self.palette.make_transparent(0)
+        if background_color or background_color == 0x000000:
```
makermelissa: This might be better expressed as: ```if background_color is not None:```

adafruit_display_text/label.py
```diff
@@ -168,6 +175,24 @@ def color(self):
...
+    @property
+    def background_color(self):
+        """Color of the background as an RGB hex number."""
+        if not self._transparent_background:
+            return self.palette[0]
+        return None
+
+    @background_color.setter
+    def background_color(self, new_color):
+        if new_color or new_color == 0x000000:
```
makermelissa: same with this: `if background_color is not None:`

makermelissa:
Thank you for making these changes. We want your code to be in its best possible form.

FoamyGuy:
@makermelissa  there is a new commit now that changes both if statements to use the `is not None` syntax. I don't foresee any issues with it, but I won't be able to test the new commit specifically on a device until later tonight.

I really appreciate everyone taking the time to point out ways it can be improved. I cut my teeth as a solo programmer, and even now I work on a team of only 2-3 developers. So feedback on source code is something I haven't got much experience with before getting involved with Circuit Python. I appreciate the feedback because I can hopefully learn and improve from it.

caternuson:
@FoamyGuy Thanks for making all the changes.

@makermelissa At this point - looks OK to me.

PR #42, opened 2020-05-22
Create getter/setter for font

kmatch98:
The label class had the "font" exposed as a variable without an explicit getter/setter.  This lead to unexpected results, such as only part of the label getting redrawn with the new font, and some of the old font remaining.  That is, some of the characters would be using the "old" font and only the changed characters would use the "new" font.  When changing the font with label.font=newFont, and calling for a text update, the drawing is only be updated for characters that were changed.  

In this pull request, I am adding an explicit definition of getter and setter for the font.  This added getter/setter causes the complete text string to be re-rendered from scratch.  By setting `self._text = ''`, then the call to `self._update_text(old_text)` will execute a full redraw of the text with the updated font.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -235,13 +235,14 @@ class Label(displayio.Group):
         return self._font
 
     @font.setter
-    def font(self, newFont):
-        old_text = self._text
-        self._text = ""
-        self._font = newFont
+    def font(self, new_font):
+        old_text=self._text
+        self._text=''
+        self._font=new_font
         bounds = self._font.get_bounding_box()
-        self.height = bounds[1]
+        self.height = bounds[1] 
         self._update_text(str(old_text))
+
 
     @property
     def anchor_point(self):
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -231,7 +231,7 @@ class Label(displayio.Group):
 
     @property
     def font(self):
-        """Font to use for text."""
+        """Font to use for text display."""
         return self._font
 
     @font.setter
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -32,7 +32,6 @@ Implementation Notes
   https://github.com/adafruit/circuitpython/releases
 """
 
-print('loading label.py Yay!')
 
 import displayio
 
@@ -110,7 +109,6 @@ class Label(displayio.Group):
             )
             / 2
         )
-        # print("y offset from baseline", y_offset)
         left = right = top = bottom = 0
         for character in new_text:
             if character == "\n":
@@ -163,7 +161,7 @@ class Label(displayio.Group):
 
             x += glyph.shift_x
 
-            # TODO skip this for control sequences or non-printables.
+            # TODO skip this for control sequences or non-qables.
             i += 1
             old_c += 1
             # skip all non-prinables in the old string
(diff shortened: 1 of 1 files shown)


kmatch98:
Please forgive me, as this is my first pull request for Adafruit libraries.  I read through the learn guide, but it wasn't obvious to me how to overcome this error when creating the documentation.  I Received the following error when creating the documents ("Sphinx"):

`/home/runner/work/Adafruit_CircuitPython_Display_Text/Adafruit_CircuitPython_Display_Text/adafruit_display_text/label.py:docstring of adafruit_display_text.label:10:Unexpected indentation.`

I'm unsure how to debug and eliminate this error message, so your suggestions and guidance will be helpful.

ladyada:
@kattni can help you get it passing next week :)

adafruit_display_text/label.py
```diff
@@ -50,7 +43,6 @@ class Label(displayio.Group):
        properties will be the left edge of the bounding box, and in the center of a M
        glyph (if its one line), or the (number of lines * linespacing + M)/2. That is,
        it will try to have it be center-left as close as possible.
-
```
FoamyGuy, line 53: I think the issue might be caused by the removal of this blank line.  This SO Q/A mentions the same error: https://stackoverflow.com/questions/45880348/how-to-remove-the-cause-of-an-unexpected-indentation-warning-when-generating-cod

kmatch98: Thanks for the suggestion, I added this blank line back in and now pylint is throwing this error:
`adafruit_display_text/label.py:46:0: C0303: Trailing whitespace (trailing-whitespace)`

Sorry if this is simple stuff, but I'm totally unfamiliar with pylint or Sphinx, so I will appreciate your guidance.

FoamyGuy: No worries. I am pretty new to sphinx and pylint myself. They can definitely be finicky.

I think the issue now is the empty line has a tab indention on it. Pylint wants it to be an empty newline with no tabs or spaces in it.

adafruit_display_text/label.py
```diff
@@ -22,21 +22,14 @@
...
 Implementation Notes
 --------------------
-
 **Hardware:**
-
 **Software and Dependencies:**
-
 * Adafruit CircuitPython firmware for the supported boards:
   https://github.com/adafruit/circuitpython/releases
-
```
FoamyGuy, line 39: It wants to have all of these empty lines between things in this comment as well.

FoamyGuy: If you match the spacing from the existing file here: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/f26a881300bd0a8ecb0038fac1991c3ecd38f3a7/adafruit_display_text/label.py#L23 it should pass I think.

kmatch98: Yay!  Thanks @FoamyGuy.  This part seems harder than figuring out the logical coding part!  Thanks for your help.

kmatch98:
Thanks to @FoamyGuy the checks now pass.

If desired, I can cancel this PR and focus on the other one that I created.  It includes these changes (sorry for combining but I forgot to fork from the original source!): See [PR # 43](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/pull/43)

Let me know how you recommend to proceed.  Thanks for all your support!

tannewt:
Closing in favor of #43

PR #47, opened 2020-05-29
Updated bounding_box "right" value to include trailing whitespace

kmatch98:
The problem:  Current label's bounding_box does not including the pixels for trailing whitespace (or any glyphs that rely on the font's "shift_x" parameter).  So, if you create a label and then use the bounding_box to set the next bit of text, then any spaces will seem like they weren't ever printed.

The fix: The "right" parameter for the bounding_box is updated to include any "glyph.shift_x", for example when a trailing whitespace is present.  This will ensure that any glyphs that used "shift_x" will be accounted-for in the right edge of the bounding_box.

Potential negative impacts:  Could mess up someone's label placement if they are already accounting for this missing space at the end.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -126,7 +126,7 @@ class Label(displayio.Group):
             glyph = self._font.get_glyph(ord(character))
             if not glyph:
                 continue
-            right = max(right, x + glyph.width)
+            right = max(right, x + glyph.width + glyph.shift_x)
             if y == 0:  # first line, find the Ascender height
                 top = min(top, -glyph.height + y_offset)
             bottom = max(bottom, y - glyph.dy + y_offset)


FoamyGuy:
Do you have a small example script that shows the incorrect behavior? I tried it out a few different ways but it seemed to me like the right value was changing when trailing whitespace was included in the text. I'm not sure I am understanding the issue correctly though

kmatch98:
Here is an example file to highlight the impact of treatement of bounding_box of the trailing whitespace on text lines.  The BDF font file can be found here:
[Bitstream Vera Sans Roman 16 pixels](https://github.com/kmatch98/CircuitPython_textMap/blob/master/fonts/BitstreamVeraSans-Roman-16.bdf)

In the BDF font I am using, the space glyph bitmap is only one pixel wide, but it includes a `shift_x` to make the whitespace (see `DWIDTH` below).  

```
STARTCHAR space
ENCODING 32
SWIDTH 317 0
DWIDTH 5 0
BBX 1 1 0 0
BITMAP
00
ENDCHAR
```

This example writes two text strings.  The first string includes a trailing whitespace.  The next string is written at an x-position that is offset by the first string's `bounding_box` width.  With the currently-released version, the two strings are bunched together:

current_version

With the version from the existing pull request that adds in the `glyph.shift_x` the two strings include the whitespace:

updatedPR_version

```
# label example to evaluate right hand whitespace
# Kevin Matocha 30 May 2020


import board
import displayio
import time
import terminalio
import fontio
import sys
import busio
#from adafruit_st7789 import ST7789
from adafruit_ili9341 import ILI9341

#  Setup the SPI display

print('Starting the display...') # goes to serial only
displayio.release_displays()


spi = board.SPI()
tft_cs = board.D9 # arbitrary, pin not used
tft_dc = board.D10
tft_backlight = board.D12
tft_reset=board.D11

while not spi.try_lock():
    spi.configure(baudrate=32000000)
    pass
spi.unlock()

display_bus = displayio.FourWire(
    spi,
    command=tft_dc,
    chip_select=tft_cs,
    reset=tft_reset,
    baudrate=32000000,
    polarity=1,
    phase=1,
)

print('spi.frequency: {}'.format(spi.frequency))

DISPLAY_WIDTH=320
DISPLAY_HEIGHT=240

#display = ST7789(display_bus, width=240, height=240, rotation=0, rowstart=80, colstart=0)
display = ILI9341(display_bus, width=DISPLAY_WIDTH, height=DISPLAY_HEIGHT, rotation=180, auto_refresh=True)

display.show(None)

print ('Display is started')


# load all the fonts
print('loading fonts...')


myGroup = displayio.Group( max_size=10 ) # Create a group for displaying
display.show(myGroup)

fontFile='fonts/BitstreamVeraSans-Roman-16.bdf' # mainText

myFont = bitmap_font.load_font(fontFile) 

textOne='space '
textTwo='following'


textColor=0xFFFFFF
backgroundColor=0x000000


#write textOne
xStart=1
yStart=20

labelOne=label.Label(font=myFont, text=textOne, color=textColor, background_color=backgroundColor, x=xStart, y=yStart)
boundingBoxOne=labelOne.bounding_box
boxOneWidth=boundingBoxOne[2]
boxOneHeight=boundingBoxOne[3]

myGroup.append(labelOne)


# write textTwo
xStart=xStart+boxOneWidth
yStart=yStart # do not change baseline position

labelTwo=label.Label(font=myFont, text=textTwo, color=textColor, background_color=backgroundColor, x=xStart, y=yStart)

myGroup.append(labelTwo)
```

Issue #11, opened 2019-05-23
Show a better message when exceeding max_glpyhs / initial text length


caternuson:
Maybe I'm incorrectly assuming the mutability of a Label? You can reassign the text, but if it is longer than the original text, then you exceed a Group limit.

```python
Adafruit CircuitPython 4.0.0 on 2019-05-20; Adafruit PyPortal with samd51j20
>>> import terminalio
>>> from adafruit_display_text import label
>>> label = label.Label(terminalio.FONT, text="123456789")
>>> label.text = "1"
>>> label.text = "1234"
>>> label.text = "1234567"
>>> label.text = "123456789"
>>> label.text = "1234567890"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "adafruit_display_text/label.py", line 174, in text
  File "adafruit_display_text/label.py", line 119, in _update_text
RuntimeError: Group full
>>>
```

iraytrace:
Given the nature of the Label, I would suggest this should either truncate the input text in the property setter method, or otherwise have a more graceful recovery.

anecdata:
I believe this behavior is currently as-designed. Label needs the maximum number of characters, or pixel width using the bounding box using the current font. Enlarging the allocation, auto-truncation option, even a scrolling text option would be great. This gets tricky with multi-line text though.

makermelissa:
If you pass in `max_glyphs=30` or whatever number is your largest size, then it will handle this correctly. If not specified, it assumes the length of the default text is the max_glyphs. I believe this issue is more about having it degrade with a better message since "Group full" is a little misleading.

dhalbert:
I just encountered this too. I'd suggest either it just truncates or we catch the "Group full" exception and throw with a more meaningful exception message like "Text too long" or "Text exceeds max_glyphs".

caternuson:
Agree with above two comments - this issue is just a suggestion for better fail behavior. And once implemented, update the displayio guide.

Would truncating be the more user friendly option?

kevinjwalters:
I just got confused with this too. I agree with @dhalbert and @jepler in Discord, a re-throw with a clearer exception would be a good thing.

kevinjwalters:
`max_glyphs` is also problematic as this is not documented in enough detail for the user to know that this needs to be `maximum(all possible text lengths) + 1` - note the `+ 1`.

[...]

PR #48, opened 2020-05-31
catch group full exception and re-raise with more descriptive message

FoamyGuy:
Solves #11

This will show a more helpful message when user tries to set text to a string that is too long.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -238,9 +238,12 @@ class Label(displayio.Group):
 
     @text.setter
     def text(self, new_text):
-        current_anchored_position = self.anchored_position
-        self._update_text(str(new_text))
-        self.anchored_position = current_anchored_position
+        try:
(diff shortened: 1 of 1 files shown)


FoamyGuy:
This small snippet shows the new behavior:
```python
import terminalio
label = label.Label(terminalio.FONT, text="123")
label.text = "1234"
```
Which is now: 
```
  File "code.py", line 4, in <module>
  File "/lib/adafruit_display_text/label.py", line 246, in text
RuntimeError: Text length exceeds max_glyphs
```

kmatch98:
@FoamyGuy I looked over your code change and it looks like it will give a better notification when the max_glyphs is exceeded.  Looks good to me.

dherrada:
@FoamyGuy ooh, thanks so much for doing this. I can't tell you how annoyed I was getting when I was learning how to use this library and kept running into this issue and getting an error message that didn't really help at all.

PR #49, opened 2020-06-04
Updated background color handling for label

kmatch98:
When using imported BDF font files with proportional typefaces, the current version of label with applied background color only covers the glyphs but not the whitespace.  Also, spaces provide some errant dots.  See image:
IMG_0653

This PR creates an additional colored bitmap tileGrid as the bottom layer below the text that is the size of the bounding box around the glyph text.  The following three images show examples of the proposed PR with differing text.

Note: The bounding box is a simple minimum bounding box to fit the whole set of glyphs.  This means that text with no ascenders and descenders (see second image below `mere a or o`) will have a smaller bounding box than other lines with taller letters.

Possible enhancements:
* Draw a bounding box the same height for all characters, to cover the space of ascenders and descenders.  This would be important when writing multiple lines of text, since it may be obvious to the viewer that the height of background text boxes differs based on the text in them.
* To avoid excess background fill regions (see last image) print each line separately so the bounding boxes will not extend past the right edge of the text.  

IMG_0659
IMG_0660
IMG_0664

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -319,7 +319,6 @@ class Label(displayio.Group):
         current_anchored_position = self.anchored_position
         self._text = ""
         self._font = new_font
-        # bounds = self._font.get_bounding_box()
         self.height = self._font.get_bounding_box()[1]
         self._update_text(str(old_text))
         self.anchored_position = current_anchored_position
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -125,7 +125,7 @@ class Label(displayio.Group):
 
     def _update_background_color(self, new_color):
 
-        if new_color == None:
+        if new_color is None:
             self._background_palette.make_transparent(0)
         else:
             self._background_palette.make_opaque(0)
@@ -228,13 +228,15 @@ class Label(displayio.Group):
         else:  # draw a "loose" bounding box to include any ascenders/descenders.
 
             # check a few glyphs for maximum ascender and descender height
-            # Enhancement: it would be preferred to access the font "FONT_ASCENT" and "FONT_DESCENT" in the imported BDF file
-            glyphs = "M j'"  # choose glyphs with highest ascender and lowest descender, will depend upon font used
+            # Enhancement: it would be preferred to access the font "FONT_ASCENT" and 
+            # "FONT_DESCENT" in the imported BDF file
+            glyphs = "M j'" # choose glyphs with highest ascender and lowest 
+                            # descender, will depend upon font used
             ascender_max = descender_max = 0
             for char in glyphs:
-                thisGlyph = self._font.get_glyph(ord(char))
-                ascender_max = max(ascender_max, thisGlyph.height + thisGlyph.dy)
-                descender_max = max(descender_max, -thisGlyph.dy)
+                this_glyph = self._font.get_glyph(ord(char))
+                ascender_max = max(ascender_max, this_glyph.height + this_glyph.dy)
+                descender_max = max(descender_max, -this_glyph.dy)
 
             box_width = self._boundingbox[2] + self._padding_left + self._padding_right
             x_box_offset = -self._padding_left
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -99,8 +99,8 @@ class Label(displayio.Group):
 
         self._background_color = background_color
 
-        bounds = self._font.get_bounding_box()
-        self.height = bounds[1]
+        #bounds = self._font.get_bounding_box()
+        self.height = self._font.get_bounding_box()[1]
         self._line_spacing = line_spacing
         self._boundingbox = None
 
@@ -115,7 +115,6 @@ class Label(displayio.Group):
             )
         )  # initialize with a blank tilegrid placeholder for background
 
-        self._padding = (padding_left, padding_top, padding_right, padding_bottom)
         self._padding_top = padding_top
         self._padding_bottom = padding_bottom
         self._padding_left = padding_left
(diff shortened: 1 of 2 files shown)


adafruit_display_text/label.py
```diff
@@ -188,6 +188,18 @@ def _update_text(self, new_text):  # pylint: disable=too-many-locals
...
 
+        background_bitmap = displayio.Bitmap(
+            self._boundingbox[2], self._boundingbox[3], 1
+        )
+
+        tile_grid = displayio.TileGrid(
+            background_bitmap, pixel_shader=self.palette, x=left, y=top
+        )
+
+        self.insert(
```
jepler, line 199: It _appears_ that every time the text is updated, an additional background TileGrid will be inserted in the group.  Did your tests cover the case where the label's text is repeatedly updated?

FoamyGuy: My test did not, but I can try that out later tonight.

kmatch98: Thanks for catching this.  I refigured the intialization so that it always adds the first element to be the background bitmap.  Now the text character TileGrids now start at element [1].

kmatch98:
Thanks for the feedback.  I eliminated the chance for possible multiple `insert` issue identified by @jepler .

Per @FoamyGuy request I added options for padding:
- If `padding_tight = True`, then the bounding box is the minimum to cover the text.  That is, the box size will depend upon what characters are inside the box.  Any of the `padding_*` variables are ignored.
- If `padding_tight = False` then the bounding box height will cover all ascenders and descenders of the current font.  Also, variable padding can be added or subtracted to this padding using the following variables: `padding_top`, `padding_bottom`, `padding_left`, `padding_right`.

Examples:
`padding_tight = True`
background_tight=True


`padding_tight = False` and no additional padding
background_tight=False, no padding


`padding_tight = False` and 5 pixels of additional padding on all four sides
background_tight=False, 5 pixel padding

adafruit_display_text/label.py
```diff
@@ -87,28 +93,48 @@ def __init__(
...
-            self.palette[0] = background_color
-            self.palette.make_opaque(0)
-            self._transparent_background = False
-        else:
-            self.palette[0] = 0
-            self.palette.make_transparent(0)
-            self._transparent_background = True
+        self.palette[0] = 0
+        self.palette.make_transparent(0)
+        self._transparent_background = True
```
jepler, line 98: Can _transparent_background be removed?  I only see it being set, and never used.

kmatch98: Deleted.  Thanks for catching.

kmatch98:
@FoamyGuy the new example added uses a separately-connected display.  If you have a chance, please verify it with your simpler code for starting the display for devices with built-in.  (Currently my daughter is using our pyPortal full-time for her studies, so I can't really verify.)  Feel free to change the example to the simpler version, since that will be easier for users to understand.

jepler:
Thanks!  This looks good at a human level but note that pylint has something to tell you.  Find where "Build CI / test" is failing, and look at the Details.  let us know if you get stuck, the formatting checks can be very ... specific.

jepler:
> adafruit_display_text/label.py:64:4: R0914: Too many local variables (16/15) (too-many-locals)
If there are any pylint diagnostics that are too onerous to deal with, they can be ignored.  Let me know if you need a pointer to docs on how to do this.

adafruit_display_text/label.py
```diff
@@ -87,28 +93,98 @@ def __init__(
...
-            self._transparent_background = True
+        self.palette[0] = 0
+        self.palette.make_transparent(0)
         self.palette[1] = color
 
-        bounds = self._font.get_bounding_box()
-        self.height = bounds[1]
+        self._background_color = background_color
+
+        # bounds = self._font.get_bounding_box()
```
jepler, line 102: Is there a reason to leave this commented-out line of code here?

kmatch98: @jepler  I'm working through the pylint errors and refigure the code.  I wanted to make sure my changes did the trick before cleaning up.  Give me a a little bit and I'll wrap up the fixes for the `label.py` file.  Also pylint is flagging several issues on the new example.  It'll take some time to get through that.

kmatch98:
I resolved all the error messages.  Let me know if you see any other changes required.

FoamyGuy:
I successfully tested the new library and example code (slightly modified) on PyPortal. 

The functionality looks good to me. 

That comment on line 322 looks to be present still. If it's unneeded we can remove it. I'm not sure I understand fully what it does right now thought. I can tinker with it a bit in the morning perhaps.

adafruit_display_text/label.py
```diff
@@ -188,6 +268,50 @@ def _update_text(self, new_text):  # pylint: disable=too-many-locals
         self._text = new_text
         self._boundingbox = (left, top, left + right, bottom - top)
 
+        # if self._background_tight:  # draw a tight bounding box
+        #     box_width = self._boundingbox[2]
+        #     box_height = self._boundingbox[3]
+        #     x_box_offset = 0
+        #     y_box_offset = top
+
+        # else:  # draw a "loose" bounding box to include any ascenders/descenders.
```
jepler: let's delete this commented-out code unless it's there for a reason; in which case... the comment needs a comment?

kmatch98: Hmm.. I'm not seeing what you're seeing with my latest revisions.  Did you grab the latest commit?  Or maybe I did something wrong.

kmatch98:
I don't see what you mean about line 322.  I pushed a few commits just now.  @jepler also said he's seeing a bunch of comments still in the file.  I checked but I don't see anything in the latest revision.  Let me know if I'm missing something.

> That comment on line 322 looks to be present still. If it's unneeded we can remove it. I'm not sure I understand fully what it does right now thought. I can tinker with it a bit in the morning perhaps.

adafruit_display_text/label.py
FoamyGuy: I think i saw this happen too. The view file page doesnt always update itself upon new commits. There is a manual refresh button but its easy to miss.

I saw the comments jepler is talking about, but i also saw they were gone after i used that manual refresh.

FoamyGuy:
In the font setter. This line:

`# bounds = self._font.get_bounding_box()`

kmatch98:
> In the font setter. This line:
> 
> `# bounds = self._font.get_bounding_box()`

Squashed it.  Dunno how I overlooked it.  Thanks for the find.  I just committed the update and it passed the checks.  Thanks for all of your help and patience!  Hope this will be of help to someone out there.  After a couple of days of this I'm ready to move on to something else...

adafruit_display_text/label.py
kmatch98: @jepler I think you reviewed an older version.  Thanks to @FoamyGuy  I removed all the extraneous comments.

FoamyGuy:
@kmatch98 Thank you for sticking through all the changes. It's really nice to have the background color working correctly with BDF fonts. And I really like the new padding functionality!

kmatch98:
Thanks guys for helping me through this and suggesting all the improvements.  Thank y’all. Maybe we’ll see more text based projects on the show n tell!

dherrada:
@kmatch98 Yeah, to echo what @FoamyGuy said this is such an awesome change. Thanks for tackling it.

Issue #58, opened 2020-07-02
Remove usage of 0-width/height bitmaps and tilegrids


makermelissa:
I discovered this when working on Blinka Displayio and looking at the circuitpython code as well, it appears a divide by zero error is happening in the background, but it's silently recovering from that. In order to avoid that, we'd like to add warnings if 0-width/height bitmaps are used, but it should be removed from here first as that would break this library.

It's being used here: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/master/adafruit_display_text/label.py#L113

I believe it's also using 0-width or height later in the code as well, though I can't pinpoint the exact location.

I put a workaround in place for Blinka Displayio for now. See https://github.com/adafruit/Adafruit_Blinka_Displayio/pull/27#pullrequestreview-441060961 for more info.

tannewt:
@kmatch98 Any thoughts? I think you added the zero-width bitmap. Is it just setting a background color?

FoamyGuy:
I believe it is only for setting the background color. The way it works now was introduced because the original `background_color` functionality was working with `terminalio.FONT` but not working very well with custom fonts. #49 has more details.

I think it's set to 0 width initially because the label may not have any text yet at this point.

It looks like the bitmap and tilegrid get created inside of `_create_background_box()` with the box sizes here: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/225b0b176afcdd712957fac504ad86ebd715bd4a/adafruit_display_text/label.py#L167

It may be possible to get rid of the placeholder 0 sized bitmap and Tilegrid in the `__init__()` without causing any issues.

But I think inside of `_create_background_box()` it could still come up as 0 if there is no text in the label and no padding. We may need to add logic to handle the "no text" case separately to avoid making the zero size bitmap.

kmatch98:
I just issued a pull request to alleviate this to prevent the zero-sized bitmap.  

I also observed some strange position changes when the text is changed after a label is created.  I will create a separate issue for that.

kmatch98:
@ foamyguy I didn't add a check for zero size of the box_width and box_height.  Do you think that will be necessary?  If so, what exception should it provide?

FoamyGuy:
@kmatch98 I do think it will be necessary. 

[...]

PR #59, opened 2020-07-02
Bitmap zero - Creates 1-pixel bitmap during initialization of tilegrid for background of label

kmatch98:
Resolving zero-sized bitmap identified in issue #58.

To include a colored background for text, a single bitmap (in a TileGrid) is used as the first element of the display group.  The first item is this background bitmap, and the remaining items are TileGrid elements, each with one character of text.

To overcome the zero-sized bitmap, I updated the initial instancing of the TileGrid to include a non-zero sized bitmap (single pixel).  

Note: This initial instance in the group is a place holder that is later updated to the correct size after all the text is created.  An alternate approach could be to create all the text TileGrids, then create the bitmap and insert it into the first position of the group.  Also, do we need to check for a zero-sized `box_width` and `box_height` before setting the bitmap background?

Also, I updated the `anchored_position` getter and setter to prevent movement of the label when the text was updated (#60).  I repaired an off-by-one error due to a difference in `int()` that truncates the decimals and changed it to `round()`.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -351,9 +351,10 @@ class Label(displayio.Group):
         """Position relative to the anchor_point. Tuple containing x,y
            pixel coordinates."""
         return (
-            int(self.x + self._anchor_point[0] * self._boundingbox[2]),
-            int(self.y + self._anchor_point[1] * self._boundingbox[3]),
-        )
+            int(self.x + self._anchor_point[0] * self._boundingbox[2] * self._scale ),
+            int(self.y + self._anchor_point[1] * self._boundingbox[3] * self._scale 
+                    - (self._boundingbox[3] * self._scale)/2 )
+                )
 
     @anchored_position.setter
     def anchored_position(self, new_position):
@@ -364,6 +365,7 @@ class Label(displayio.Group):
         new_y = self.y = int(
             new_position[1]
             - self._anchor_point[1] * (self._boundingbox[3] * self._scale)
+            + (self._boundingbox[3] * self._scale)/2
         )
         self._boundingbox = (new_x, new_y, self._boundingbox[2], self._boundingbox[3])
         self.x = new_x
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -235,13 +235,14 @@ class Label(displayio.Group):
         return self._font
 
     @font.setter
-    def font(self, newFont):
-        old_text = self._text
-        self._text = ""
-        self._font = newFont
+    def font(self, new_font):
+        old_text=self._text
+        self._text=''
+        self._font=new_font
         bounds = self._font.get_bounding_box()
-        self.height = bounds[1]
+        self.height = bounds[1] 
         self._update_text(str(old_text))
+
 
     @property
     def anchor_point(self):
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -353,12 +353,10 @@ class Label(displayio.Group):
         return (
             int(
                 self.x
-                + self._boundingbox[0]
                 + self._anchor_point[0] * self._boundingbox[2]
             ),
             int(
                 self.y
-                + self._boundingbox[1]
                 + self._anchor_point[1] * self._boundingbox[3]
             ),
         )
@@ -372,7 +370,6 @@ class Label(displayio.Group):
         new_y = self.y = int(
             new_position[1]
             - self._anchor_point[1] * (self._boundingbox[3] * self._scale)
-            + (self._boundingbox[3] * self._scale) / 2
         )
         self._boundingbox = (new_x, new_y, self._boundingbox[2], self._boundingbox[3])
         self.x = new_x
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -231,7 +231,7 @@ class Label(displayio.Group):
 
     @property
     def font(self):
-        """Font to use for text."""
+        """Font to use for text display."""
         return self._font
 
     @font.setter
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -164,17 +164,14 @@ class Label(displayio.Group):
         box_width = max(0, box_width)  # remove any negative values
         box_height = max(0, box_height)  # remove any negative values
 
-        if box_width > 0 and box_height > 0:
-            background_bitmap = displayio.Bitmap(box_width, box_height, 1)
-            tile_grid = displayio.TileGrid(
-                background_bitmap,
-                pixel_shader=self._background_palette,
-                x=left + x_box_offset,
-                y=y_box_offset,
-            )
-            return tile_grid
-        else:
-            return None
+        background_bitmap = displayio.Bitmap(box_width, box_height, 1)
+        tile_grid = displayio.TileGrid(
+            background_bitmap,
+            pixel_shader=self._background_palette,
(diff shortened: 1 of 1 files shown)


adafruit_display_text/label.py
```diff
@@ -372,7 +370,6 @@ def anchored_position(self, new_position):
         new_y = self.y = int(
             new_position[1]
             - self._anchor_point[1] * (self._boundingbox[3] * self._scale)
-            + (self._boundingbox[3] * self._scale) / 2
```
FoamyGuy, line 375: I think we need to keep this line. It accounts for the fact that by default the y position of the label is relative to the vertical center on the label rather than the top of the label.

kmatch98: Thanks for the suggestion.  I think am often confused because the initial x,y position of the label is defined so differently when creating the label and then switching to anchored position mode. For the case of changing the text in a label, I get the current anchored position, reset the text and then set the anchor position again. This is to accommodate if the label changes size. 

Imagine if I set the text to the same value, then when you use the getter, change the text and then apply the setter, then the x and y values should be exactly the same. Said another way, the getter should just be the inverse of the setter. So if the setter and getter aren’t inverses then something is wrong.  This difference is what is causing the text to walk around. 

Let me know your thoughts and then I can try a few cases and see how to get it to work.

FoamyGuy: Ah It looks like I just forgot to update the getter when I did the setter recently to make it work with scaled labels. 

I think we can put this line back in the setter and then put the inverse of it in the getter in there it will be minus instead of plus, like this:
```python
- (self._boundingbox[3] * self._scale) / 2
```

We need to add the `_scale` factor multiplication in on both axis in the getter as well.

kmatch98: **anchor_position getter/setter:** Thanks for the clarification.  Got it, I just updated, tested and committed another revision.  I added the scale factor into the getter, but I haven't verified any of the scale options.

**Bitmap handling:** I'd like to consider redoing the bitmap sections in `_update_text` and `_create_background_box`.  But to do this, we will need to add a `self._background_present` state variable to the class to determine if a background is present or not.  That will allow the code to know whether to insert, overwrite or delete the background into the group depending upon any changes that are requested.  Seems a bit overcomplicated but I can't think of another way.  Your suggestions are appreciated!

makermelissa: Rather than a background_present variable, why not just use `self._background_color is not None`?

FoamyGuy: I think `self._background_color` could be set to a color instead of `None`, but we could still get in a situation where we need to have `self._background_present` set to `False`. This would cover the case when the current width of the label is 0 pixels due to having no text in it. The background color can be set to something. But currently we don't want to draw the Bitmap because it would have been 0 pixels.

FoamyGuy:
The `anchored_position` parts look good to me now. I will test them out a bit today as well.

I was thinking about the _create_background_box() 0 size issue a bit more. Instead of making a new property `_background_present` what if we use an if statement inside of [_update_text() here](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/2d65dc0f5ca8b21c2d6b66119948599ffd3e67c2/adafruit_display_text/label.py#L270) like this:
```python
if len(new_text) + self._padding_left + self._padding_right > 0:
    self[0] = self._create_background_box(lines, y_offset)
```
That way we will skip making the background all together if there is nothing to draw.

FoamyGuy:
After testing this out a little bit it does seem to resolve the text getting moved problem. 

I also tried out an if statement like above, however I also noticed that there are issues when trying to change the background color to None. So I think it needs some more tweaking still. 

I think solution can also help solve the recently the recently created #61 because I think we can get it to avoid creating/drawing the bitmap if it's not needed.

FoamyGuy:
I never did get it working quite right in all scenarios and my attempts started to feel a bit over-complicated. I will come back to it a bit later tonight, or tomorrow with fresh eyes.

FoamyGuy:
I have a branch here: https://github.com/FoamyGuy/Adafruit_CircuitPython_Display_Text/blob/remove_0_width_bitmap/adafruit_display_text/label.py that I've been using to work on this. This is working for changing background colors, and setting background color to None. I think it should avoid usage of 0 size bitmaps. But there is still some problem with it because the background box size is incorrect. Also there are lots of debugging prints in it right now. 

You can check out this branch and poke around a bit if you like. I will come back to it tomorrow to try to resolve the remaining issues with it.

kevinjwalters:
How does the bitmap background differ from setting the background before? I set it to blue in the application with screenshow visible at top of https://learn.adafruit.com/clue-sensor-plotter-circuitpython/circuitpython-sensor-plotter and that was before the bitmap?

FoamyGuy:
@kevinjwalters the code you linked is using the default font `terminalio.FONT` which was always working working correctly. 

The issue was with custom BDF Fonts. On those the background color was not covering the whole label, but instead each individual glyph. See #49 for more info about that issue. The bitmap background was introduced to allow backgrounds to work with custom fonts.

kevinjwalters:
@FoamyGuy Thanks, I had been puzzling over this one!

kmatch98:
@FoamyGuy I reviewed the code and ran this version that you are working on (https://github.com/FoamyGuy/Adafruit_CircuitPython_Display_Text/commit/51e13b2e619c9069f74c60eeb4c2f9f64b052247).

__Sizing:__ The background sizing looks right to me, I tried it with 0 padding and with +4 padding.

__Logic and function organization:__ I agree with how you added a state variable `self._added_background_tilegrid` to keep track of the presence of the background.  With this addition, you can now delete the initial creation of the bitmap (line 116 `if self._background_color:...`).  The work of adding/removing the background can now fully fall to the combined efforts of `_update_background_color` and `_create_background_box` functions.

Please consider that if the background gets switched to `None` to just pop the background tileGrid out of the Group, and set `self._added_background_tilegrid = False`.  That will reduce the memory usage by a little bit and I suspect it will put a little less work on the display to deal with the transparent bitmap.  Also, the same "remove bitmap" function should be used if the width or height of the background is somehow <= 0.

Overall, adding this state variable adds one more state variable, but I think it will help resolve the issues that were raised, including the zero-sized bitmap and the option to reduce memory usage without any bitmap.

Once you read these comments, let me know if you want another checkout of the code, or if you want me to make any of these changes.  Keep up the good work, and thanks for everyone fo finding these issues!
------

> I have a branch here: https://github.com/FoamyGuy/Adafruit_CircuitPython_Display_Text/blob/remove_0_width_bitmap/adafruit_display_text/label.py that I've been using to work on this. This is working for changing background colors, and setting background color to None. I think it should avoid usage of 0 size bitmaps. But there is still some problem with it because the background box size is incorrect. Also there are lots of debugging prints in it right now.
> 
> You can check out this branch and poke around a bit if you like. I will come back to it tomorrow to try to resolve the remaining issues with it.

FoamyGuy:
Thanks @kmatch98. That all sounds good to me. 

My intention is to work on finishing this up later today. If you have some time today you can work on some of those last few tweaks mentioned if you want. But if not no worries.

Once I get to it I'll push to that same branch with an open PR on your fork. That way when it's ready you can merge that PR and everything should show up here I believe.

kmatch98:
@foamyguy I’ll be tied up today until tonight but let me know if there are any loose ends to deal with then.

kmatch98:
Thank you for your support on this!

PR #62, opened 2020-07-07
fix anchored position getter and setter

FoamyGuy:
I believe these changes should fix issues related to the label getting moved to unexpected places when using `anchored_position` getter and setters.


FoamyGuy:
Closing this in favor of #59 which includes the same fix.

PR #74, opened 2020-07-23
dont create a background bitmap if background_color is None

### LINKED ISSUE #72: Don't allocate a background bitmap when background_color is None
I'm getting this even when background_color is None.

```
  File "code.py", line 50, in <module>
  File "code.py", line 47, in <module>
  File "adafruit_pyportal.py", line 1028, in fetch
  File "adafruit_pyportal.py", line 601, in set_text
  File "adafruit_display_text/label.py", line 123, in __init__
  File "adafruit_display_text/label.py", line 295, in _update_text
  File "adafruit_display_text/label.py", line 205, in _update_background_color
  File "adafruit_display_text/label.py", line 162, in _create_background_box
MemoryError: memory allocation failed, allocating 3780 bytes
```

FoamyGuy:
I think this will resolve #72

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -292,7 +292,7 @@ class Label(displayio.Group):
         self._text = new_text
         self._boundingbox = (left, top, left + right, bottom - top)
 
-        if self.background_color:
+        if self.background_color is not None:
             self._update_background_color(self._background_color)
 
     @property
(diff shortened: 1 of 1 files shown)


adafruit_display_text/label.py
```diff
@@ -292,7 +292,8 @@ def _update_text(
         self._text = new_text
         self._boundingbox = (left, top, left + right, bottom - top)
 
-        self._update_background_color(self._background_color)
+        if self.background_color:
```
tannewt, line 295: Can background color be 0x000000 for black? That will fail this test. I think you want:
```suggestion
        if self.background_color is not None:
```

(I just ran into this with requests and a page of 0 content length.)

FoamyGuy: Whoops. Thank you pointing that out. The new commit changes to use `is not None`

PR #79, opened 2020-08-07
Add `bitmap_label.py` library for reduced memory usage for displaying text

kmatch98:
I created this new `bitmap_label.py` library as a drop-in replacement for `label.py` but it uses less memory.

This new library uses a bitmap to store text.  It has been calibrated against the position of `label.py`.

I include two examples that can be used to compare the two versions.  The new library consistently uses less memory than the equivalent text in `label.py`.  This can be evaluated directly using the example file `display_text_label_vs_bitmap_label_simpletest.py` which also displays the amount of memory utilized.

#### Other impacts:
The speed of using the `bitmap_label` library is inferior compared to the existing `label` library, primarily limited by the current of copying the bitmap glyphs.  This bitmap copy performance is the subject of [a related issue with display.io](https://github.com/adafruit/circuitpython/issues/3004).


kmatch98:
I received this error message from black.  I ran black locally and it had no issues.  

I appreciate if anyone can point me in the right direction to help resolve this formatting issue.

image

kmatch98:
I commited some updates to resolve some of the pylint errors on `bitmap_label` but over-rode some of them.  Also, the examples retain some pylint errors.  Please let me know whether I need to resolve these.

adafruit_display_text/bitmap_label.py
```diff
@@ -0,0 +1,445 @@
...
+"""
+
+import displayio
+
+__version__ = "0.0.0-auto.0"
+__repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
+
+
+class Label(displayio.Group):
+    """A label displaying a string of text. The origin point set by ``x`` and ``y``
```
FoamyGuy, line 59: In this comment, lets note the fact that this implementation uses a single bitmap internally as opposed to the multi-tilegrid approach of the other label class.

adafruit_display_text/bitmap_label.py
```diff
@@ -0,0 +1,445 @@
...
+       """
+
+    # pylint: disable=unused-argument, too-many-instance-attributes, too-many-locals, too-many-arguments
+    # Note: max_glyphs parameter is unnecessary, this is used for direct
+    # compatibility with label.py
+
+    # Class variable
+    # To save memory, set Label._memory_saver=True and avoid storing the text string in the class.
+    # If set to False, the class saves the text string for future reference.
+    _memory_saver = True
```
FoamyGuy, line 93: I think we should provide access to this setting as a named parameter in the `__init__()`. As it is now it would be difficult to change in user code I think. 

I would suggest renaming it to to `write_only` as well to make it clear to the user that if they set it to True then they won't be able to read the current text value. 

I think it may be best to default it to `False` as well. So we match the read/write ability of existing label.py.

examples/display_text_label_vs_bitmap_label_simpletest.py
```diff
@@ -0,0 +1,180 @@
+# Sample for comparing label and bitmap_label memory usage with Builtin or loaded BDF fonts
```
FoamyGuy, line 1: I think we should either add one more example `display_text_bitmap_label_simpletest.py` or else rename this one and make it focus on just showing simple usage of the bitmap label.

I think it's okay to drop down to just the more advanced comparison script and turn this one into the simpletest. 

Or if you'd prefer to keep both comparisons we can create a new one for the simpletest.

examples/display_text_label_vs_bitmap_label_alternating.py
```diff
@@ -0,0 +1,370 @@
...
+
+
+print("***")
+
+display.auto_refresh = True
+
+while True:
+    print("bitmap_label")
+    time.sleep(0.1)
+    display.show(bmap_group)
```
FoamyGuy, line 363: What do you think about showing both the label and the bitmap_label on the screen at the same time instead of blinking back and forth over time? If we do that we could change `alternating` to `advanced`

Also possibly changing up the text in one of the labels to include some informative text like memory used. So the user can see at a glance on the screen the difference and change out the script with their own strings to observe results.

adafruit_display_text/bitmap_label.py
kmatch98: Good point.  Agreed that `False` is the best default case.

However, perhaps the use of `write_only` might make the user think that they can re-write new text into it.  I'm open to name changes, the only reason I called it `memory_saver` was that if people were trying to shave some bytes they might notice.  How about something like `save_text_string`?  Open to suggestions.

examples/display_text_label_vs_bitmap_label_simpletest.py
kmatch98: I agree with you about the _simpletest version.  Should it just be a copy of the `display_text_label_simpletest.py` with the library name changed?

examples/display_text_label_vs_bitmap_label_alternating.py
kmatch98: I made the alternating demo so that I could check the pixel-to-pixel accuracy between the two versions.  Most folks won't care about this, so having it side-by-side would be useful as you mention.  Are you willing to take on updating the examples (`_simple_test` and `_advanced`)?

FoamyGuy: Yep I can tackle the example scripts this evening.

examples/display_text_label_vs_bitmap_label_simpletest.py
FoamyGuy: Yep, agree with that.

adafruit_display_text/bitmap_label.py
FoamyGuy: Ah yeah I didn't think the `write_only` through all the way last night, I can see how that would be confusing as well. I do like `save_text_string`. We can mention in the docstring for it that it's primary use is for saving memory, so hopefully people who go looking for it will still find it. 

Yep I agree as well with returning `None` from the `text` getter when this option is enabled, definitely do not want it to raise an exception I don't think.

adafruit_display_text/bitmap_label.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+* Author(s): Kevin Matocha
+
+Implementation Notes
+--------------------
+
+**Hardware:**
+
+.. todo:: Add links to any specific hardware product page(s), or category page(s). Use
```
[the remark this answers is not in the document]
kmatch98: Deleted.

adafruit_display_text/bitmap_label.py
kmatch98: Updated the implementation details and added a few limitations.

adafruit_display_text/bitmap_label.py
```diff
@@ -0,0 +1,445 @@
...
+unordered list & hyperlink rST
+   inline format: "* `Link Text <url>`_"
+
+**Software and Dependencies:**
+
+* Adafruit CircuitPython firmware for the supported boards:
+  https://github.com/adafruit/circuitpython/releases
+
+.. todo:: Uncomment or remove the Bus Device and/or the Register library dependencies based
```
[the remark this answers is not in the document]
kmatch98: I deleted these lines.

kmatch98:
I updated the getters and setters to match `label.py`.  

I have a couple of pylint errors that I'm not sure how to resolve.  These are three setters (font, text, line_spacing) that only return errors.  I suspect it is flagging these functions because they don't use the input variables or any references to `self`.  Any suggestions on how to solve this?

Here is the code I used to verify the operation of the getters/setters:
```
gc.collect()
label_start_memory = gc.mem_free()
start_time = time.monotonic()

bmap_label = label.Label(
    font=terminalio.FONT,
    text=my_string,
    color=0xFFFFFF,
    max_glyphs=len(my_string),
    background_color=0xFF0000,
    padding_bottom=0,
    padding_left=0,
    padding_right=0,
    padding_top=0,
    background_tight=False,
    x=10,
    y=30,
    line_spacing=1.25,
    scale=my_scale,
)
end_time = time.monotonic()

gc.collect()
label_end_memory = gc.mem_free()

bmap_group = displayio.Group(max_size=1)  # Create a group for displaying
bmap_group.append(bmap_label)


print("***")
print("{} memory used: {}".format(version, label_start_memory - label_end_memory))
print("{} time to process: {} seconds".format(version, end_time - start_time))
display.auto_refresh = True

display.show(bmap_group)

print("bounding_box: {}".format(bmap_label.bounding_box))
import time
time.sleep(1)
print('color: {}'.format(bmap_label.color))
bmap_label.color=0xFFFF00
time.sleep(1)
print('background_color: {}'.format(bmap_label.background_color))
bmap_label.background_color=0x00FFFF
time.sleep(1)
print('background_color: {}'.format(bmap_label.background_color))
bmap_label.background_color=0x0000FF
time.sleep(1)

print('text: {}'.format(bmap_label.text))
#bmap_label.text='new text' # should cause an error

print('current linespacing: {}'.format(bmap_label.line_spacing))
#bmap_label.line_spacing=3 # should cause an error
print('font: {}'.format(font))
#bmap_label.font=terminalio.FONT # should cause an error
time.sleep(1)

print('anchored_position: {}'.format(bmap_label.anchored_position))
bmap_label.anchored_position=(320/2, 240/2)
print('anchored_position: {}'.format(bmap_label.anchored_position))
time.sleep(1)

print('anchor_point: {}'.format(bmap_label.anchor_point))
bmap_label.anchor_point=(0.5,0.5)
time.sleep(1)

print('anchored_position: {}'.format(bmap_label.anchored_position))
bmap_label.anchored_position=(320/2, 240/2)
print('anchored_position: {}'.format(bmap_label.anchored_position))
time.sleep(1)

print('anchor_point: {}'.format(bmap_label.anchor_point))
bmap_label.anchor_point=(0,0)
print('anchor_point: {}'.format(bmap_label.anchor_point))
time.sleep(1)


print('anchor_point: {}'.format(bmap_label.anchor_point))
bmap_label.anchor_point=(0.5,0.5)
time.sleep(1)




while True:
    pass
```

FoamyGuy:
I think it's okay to add the pylint ignore inside of those three functions to disable the `no-self-use` check. 

Although I'm also curious how hard it would be (or if there is any downside) to allow those properties to be mutable by having them call `_place_text()` again. I'm going to try it out and see if I can get it working.

kmatch98:
Good point maybe if the bitmap is recreated then it will should work fine. Just have to be sure it releases the previous bitmap so it can get garbage collected. Keep me posted on what you find.

FoamyGuy:
I got dug into it some and I see it's going to be a bit more difficult that I was thinking to make them mutable at least for `text`. 

I think we will need to recreate the tilegrid too possibly, and do some calculations to figure out the correct size again. Which will mean that we need a few of the parameters to be stored as class variables so that they can be accessed outside of `__init__()`

I think it's a good goal to make them mutable at some point if we can, but I'm not sure it's worth holding up this PR for it. 

For now I do think it's okay to add the pylint ignores since those methods currently are only provided for API matching and to throw the error letting the user know that they aren't mutable. They don't need to use `self` for anything in order to do that.

kmatch98:
Ok, I just pushed with the `pylint: disable=no-self-use` to clear the pylint error.  
I think I covered all your previous corrections, please let me know what else you find.

#### Future: Maybe this should be an issue
I agree that making it mutable will take some time.  I've never done anything with actively freeing up instances, so that will be new ground.  But I'll spend my time working on getting the bitmap.blit function cleaned up before returning to this.

examples/display_text_label_vs_bitmap_label_simpletest.py
kmatch98: Merged your `simpletest` example.

examples/display_text_label_vs_bitmap_label_alternating.py
kmatch98: Merged your `display_text_label_vs_bitmap_label_comparison.py` example.

kmatch98:
I merged your last changes and all passed the checks.  I like the quote on the comparison example.  

I think I resolved all of your Review changes that you requested, but double check.  Otherwise, I think this is good to go.

kmatch98:
I hunted for the Strange glyph loading memory usage and remembered why.  

For determining the ascender/descender height, It checks the max dimensions of several glyphs: M, j, space, and ‘ apostrophe. So it’s best to load these four glyphs are loaded in the `load_glyphs` statement. I will update the examples tomorrow to add the apostrophe.

FoamyGuy:
I looked over the current version tonight and it looks good to me. I'll mark it approve after you get that last change in and we can see if we get any other feedback prior to merging

kmatch98:
Ok, I just updated the examples and passed the checks.  Let me know what else you find that needs updates.

#### Future enhancements:
- Make `bitmap_label` type more mutable
- If `bitmap.blit` function is merged in the CircuitPython core, then update `bitmap_label.py` to utilize this faster copy function

kmatch98:
Awesome.  Thanks for your help working on this!

Issue #80, opened 2020-08-07
y position of text can be wrong if constructor does not specify text


kevinjwalters:
I updated my libraries last night to ones from `adafruit-circuitpython-bundle-5.x-mpy-20200806.zip`. I noticed the performance improvement with moving around text but I also spotted one solitary piece of text that was misplaced slightly off screen.

I've just made a simple recreation of this, I think it's when `text` is not specified or it is empty string, `y` is set, and then `text` is set. I am using scaled text here.

```
Press any key to enter the REPL. Use CTRL-D to reload.
Adafruit CircuitPython 5.3.0 on 2020-04-29; Adafruit CLUE nRF52840 Express with nRF52840
>>>
>>> import displayio, terminalio, board
>>> from adafruit_display_text.label import Label
>>> fdim = terminalio.FONT.get_bounding_box()
>>> l1 = Label(terminalio.FONT, max_glyphs=9, scale=3, color=0xff0000)
>>> l2 = Label(terminalio.FONT, text="", max_glyphs=9, scale=3, color=0x00ff00)
>>> l3 = Label(terminalio.FONT, text="Chad", max_glyphs=9, scale=3, color=0x0000ff)
>>> l1.y = l2.y = l3.y = round(240 - (3 * fdim[1] / 2))
>>> (l1.y, l2.y, l3.y)
(219, 219, 219)
>>> l2.x = 80
>>> l3.x = 160
>>> tg = displayio.Group(max_size=3)
>>> tg.append(l1)
>>> tg.append(l2)
>>> tg.append(l3)
>>> board.DISPLAY.show(tg)
>>> l1.text="Wot"
>>> l2.text="No"
>>> (l1.y, l2.y, l3.y)
(240, 240, 219)
```

Only `l3` is correctly position, `l1` and `l2` are half off screen.

I see #78 which sounds like it could be in the same area?

Setting to whitespace seems to be a workaround for me, e.g. `text=" "`, and might be able to re-order my code to set `text` once in constructor.

kmatch98:
Put this in my queue.  I think it’s probably different than the case you mention. My first thought is that the `bounding_box` has something wrong with it but I’ll have to dig in. Glad you found a workaround in the meantime.

kmatch98:
@kevinjwalters thanks for the heads-up on this issue with the `text` setter.  

[...]

PR #81, opened 2020-08-08
Updated label.py to correct position of initial blank text

kmatch98:
This is to resolve issue https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/80

Some background on label positioning:
There are two ways to specify a label's position, either by `x`,`y` or by `anchor_point` and `anchored_position`.
These two methods could be inconsistent, so I coded this so that setting `anchored_position` actually updates just the x,y locations.  That way we don't have inconsistencies betwee the two.  And the `anchored_position` getter just calculates the value based on the stored `x,y` value.  
However, that means that the positioning of the label will depend upon the order of setting the position.  For example, if you set `x,y` directly, then it will go to that position until you set `anchored_position` to another value.  

In order to keep the `anchored_position` consistent, the `text` setter first gets the current `anchored_position`, then updates the text (in case the size changes) and then resets the `anchored_position` (which sets the appropriate x,y values).

To solve this issue #80, I updated `anchor_point` to allow a value of None.  That way, if the `anchor_point` is None, then the `anchored_position` is also invalid.  That will ensure that the `x,y` values are not changed when the `anchor_point` is not set and then the text is changed.  I hope this is solution is generic enough to resolve most issues like this.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -397,15 +397,14 @@ class Label(displayio.Group):
     def anchored_position(self, new_position):
         if (self._anchor_point is None) or (new_position is None):
             return  # Note: anchor_point must be set before setting anchored_position
-        else:
-            new_x = int(
-                new_position[0]
-                - self._anchor_point[0] * (self._boundingbox[2] * self._scale)
-            )
-            new_y = int(
-                new_position[1]
-                - (self._anchor_point[1] * self._boundingbox[3] * self._scale)
-                + round((self._boundingbox[3] * self._scale) / 2.0)
-            )
-            self.x = new_x
-            self.y = new_y
+        new_x = int(
+            new_position[0]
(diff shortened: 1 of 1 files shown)


adafruit_display_text/label.py
```diff
@@ -393,14 +395,17 @@ def anchored_position(self):
...
-        )
-        new_y = int(
-            new_position[1]
-            - (self._anchor_point[1] * self._boundingbox[3] * self._scale)
-            + round((self._boundingbox[3] * self._scale) / 2.0)
-        )
-        self.x = new_x
-        self.y = new_y
+        if (self._anchor_point is None) or (new_position is None):
+            pass  # Note: anchor_point must be set before setting anchored_position
```
tannewt, line 399:
```suggestion
            return  # Note: anchor_point must be set before setting anchored_position
```

If you return, then you don't need the else and can leave the existing indentation.

PR #83, opened 2020-08-15
Fix anchored position rounding

### LINKED ISSUE #82: x position gets shifted over time when changing the text beingdisplayed
This sample code illustrates the issue:
```python
import board
import time
import terminalio

text = "Helloworld"
font = bitmap_font.load_font("fonts/Helvetica-Bold-16.bdf")
text_area = label.Label(font, text=text*5)
text_area.anchor_point = (0.75,0)
text_area.anchored_position = (190, 10)

board.DISPLAY.show(text_area)

i = 0
while True:
    time.sleep(0.01)
    if i % 2 == 0:
        text_area.text = "Hello world1".format(i)
    else:
        text_area.text = "Hello\nworld {}".format(i)
    i += 1
```

The label will move itself left until it reaches the edge of the screen. 

This appears to be a rounding issue with floating point math in the anchored_position getter and setter which get called when the text is updated.

FoamyGuy:
This resolves #82. The sample script in that issue and several other variations of it that I tested now result in the label remaining in the same position over time.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -385,7 +385,7 @@ class Label(displayio.Group):
         if self._anchor_point is None:
             return None
         return (
-            int(self.x + (self._anchor_point[0] * self._boundingbox[2] * self._scale)),
+            int(self.x + round(self._anchor_point[0] * self._boundingbox[2] * self._scale)),
             int(
                 self.y
                 + (self._anchor_point[1] * self._boundingbox[3] * self._scale)
@@ -399,7 +399,7 @@ class Label(displayio.Group):
             return  # Note: anchor_point must be set before setting anchored_position
         new_x = int(
             new_position[0]
-            - self._anchor_point[0] * (self._boundingbox[2] * self._scale)
+            - round(self._anchor_point[0] * (self._boundingbox[2] * self._scale))
         )
         new_y = int(
             new_position[1]
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -385,7 +385,10 @@ class Label(displayio.Group):
         if self._anchor_point is None:
             return None
         return (
-            int(self.x + round(self._anchor_point[0] * self._boundingbox[2] * self._scale)),
+            int(
+                self.x
+                + round(self._anchor_point[0] * self._boundingbox[2] * self._scale)
+            ),
             int(
                 self.y
                 + (self._anchor_point[1] * self._boundingbox[3] * self._scale)
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -390,9 +390,11 @@ class Label(displayio.Group):
                 + round(self._anchor_point[0] * self._boundingbox[2] * self._scale)
             ),
             int(
-                self.y
-                + (self._anchor_point[1] * self._boundingbox[3] * self._scale)
-                - round((self._boundingbox[3] * self._scale) / 2.0)
+                round(
+                    self.y
+                    + (self._anchor_point[1] * self._boundingbox[3] * self._scale)
+                    - ((self._boundingbox[3] * self._scale) / 2.0)
+                )
             ),
         )
 
@@ -405,9 +407,11 @@ class Label(displayio.Group):
             - round(self._anchor_point[0] * (self._boundingbox[2] * self._scale))
         )
         new_y = int(
-            new_position[1]
-            - (self._anchor_point[1] * self._boundingbox[3] * self._scale)
-            + round((self._boundingbox[3] * self._scale) / 2.0)
+            round(
+                new_position[1]
+                - (self._anchor_point[1] * self._boundingbox[3] * self._scale)
+                + ((self._boundingbox[3] * self._scale) / 2.0)
+            )
         )
         self.x = new_x
         self.y = new_y


adafruit_display_text/label.py
```diff
@@ -399,7 +402,7 @@ def anchored_position(self, new_position):
             return  # Note: anchor_point must be set before setting anchored_position
         new_x = int(
             new_position[0]
-            - self._anchor_point[0] * (self._boundingbox[2] * self._scale)
+            - round(self._anchor_point[0] * (self._boundingbox[2] * self._scale))
         )
         new_y = int(
```
kmatch98, line 407: Here is my proposed update for the `anchored_position` setter for variable `new_y`:
```
        new_y = int(round(
            new_position[1]
            - (self._anchor_point[1] * self._boundingbox[3] * self._scale)
            + ((self._boundingbox[3] * self._scale) / 2.0)
        ))
```

adafruit_display_text/label.py
```diff
@@ -385,7 +385,10 @@ def anchored_position(self):
         if self._anchor_point is None:
             return None
         return (
-            int(self.x + (self._anchor_point[0] * self._boundingbox[2] * self._scale)),
+            int(
+                self.x
+                + round(self._anchor_point[0] * self._boundingbox[2] * self._scale)
+            ),
             int(
```
kmatch98, line 392: Here are my proposed changes for the `anchored_position` y-getter
```
            int(round(
                self.y
                + (self._anchor_point[1] * self._boundingbox[3] * self._scale)
                - ((self._boundingbox[3] * self._scale) / 2.0)
            )),
```

FoamyGuy:
Thank you @kmatch98! I'm trying out your changes now.

FoamyGuy:
I was able to use your code and reproduce the y shifting. I tested your changes successfully and it does resolve the shifting. Nice catch! Thank you. 

Latest commit adds those changes.

kmatch98:
@jepler do these fixes solve the moving text bug you observed on your Black Lives Matter Say Their Names display?

kmatch98:
Reminder: this Rounding bug needs to be corrected on `bitmap_label.py`

jepler:
No, this doesnb't fix the problem for me.  Additionally, it changed the vertical positioning of some labels in my code that were NOT using anchored_point / anchored_position.

These labels were shifted up:
```
# Create a 3 line set of text for BLM
blm_font = [None, None, None]
for line in range(3):
    label = adafruit_display_text.label.Label(
        font, color=0xFFFFFF, x=8, y=line * 84+12, max_glyphs=16,
    )
    blm_font[line] = label
    blm_group.append(label)
```

These labels "slide slowly left" as their text is changed repeatedly.  Reassigning the anchored_position and anchor_point properties keeps them in position, though
```
# Create a 2 line set of font text for names
names_font = [None, None]
for line in range(2):
    label = adafruit_display_text.label.Label(
        font,
        color=0xFFFFFF,
        max_glyphs=16,
    )
    # Center each line horizontally, position vertically
    label.anchor_point = (0.5, 0)
    label.anchored_position = (200, line*84+42)
    names_font[line] = label
```

The full program is https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/master/CircuitPython_SharpDisplay_Displayio/code.py and the font is also in the same folder on github.

jepler:
Here's a reproducer that doesn't actually need the display, but it *DOES* need the font.
```
import terminalio
import adafruit_display_text.label
import adafruit_bitmap_font.bitmap_font

font = adafruit_bitmap_font.bitmap_font.load_font("/GothamBlack-54.bdf")
font.load_glyphs(b'Abcdefghijklmnop')

label = adafruit_display_text.label.Label(
    font,
    color=0xFFFFFF,
    max_glyphs=16,
)
# Center each line horizontally, position vertically
label.anchor_point = (0.5, 0)
label.anchored_position = (200, 42)

while True:
    for i in range(16):
        label.text = "Abcdefghijklmnop"[:i]
    print(label.x, label.y)
```

FoamyGuy:
Thanks for the update jepler. The vertical position changing for some of the texts not using anchored_position definitely is odd to me. I believe those should have been unaffected since our changes were only inside of `anchored_position` getter and setter. We'll step through the logic in a bit more detail. Perhaps it's being utilized somewhere internally that I am overlooking. 

Thanks for sharing this code. We will get it tested out and once we have another attempt at a fix we can test against this project.

jepler:
I will be putting in a PR to work around this issue in the Learn repo's script, so take care to grab the script at https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/d130bff4eb53d5b25edb808c204988e88eb47557/CircuitPython_SharpDisplay_Displayio/code.py for your testing

FoamyGuy:
@jepler I ran your reproducer code on my end on a Wio Terminal device and I don't think I am seeing anything out of the ordinary. 

When I run that code and let it go for several seconds I am seeing the same x,y values printed over and over:
```
-132 76
-132 76
-132 76
-132 76
-132 76
-132 76
-132 76
-132 76
-132 76
...
```

FoamyGuy:
Also it's worth noting that when using `label.anchor_point = (0.5, 0)` it would be expected that the `x` position could change if the text in the label changes, since the `anchor_point` is at 50% the left and right edges of the bounding box will get bigger or smaller as text gets added or removed from the label. 

However that is only if the actual String within the text is different. Resetting to the same text that is already present should not result in any movement ideally.

FoamyGuy:
One more quick update for now: @jepler are you certain you copied in the edited code from this PR to your device? I reverted back to the "real" display_text library and ran your reproducer and I do see the x position that gets printed changing. 

But when I run that same code with the modified display_text library from this PR I always see `-132` printed for the x position.

I know you are working on a Sharp display of some sort, I don't have access to try with that, but I am curious what microcontroller are you using? If it is something I have access to I'd like to give this reproducer code a try on my end using the same device if possible.

kmatch98:
I ran @jepler 's [code](https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/d130bff4eb53d5b25edb808c204988e88eb47557/CircuitPython_SharpDisplay_Displayio/code.py) on my ILI9341 display (320x240) using the currently-released code for `label.py` and recreated the issue that he observed, with the text position moving (`anchored_position` is changing).  

Without changing the code, I changed the `label` library to @FoamyGuy 's updated code with the latest commit and this solved the issue, and the `anchored_position` did not change.  (Note: My screen is smaller than the one used by jepler, so it got cut off at the top, but otherwise this code worked fine for me with the library in this PR.)

Also, I found a Black Lives Matter font here: https://blacklivesmatterfont.com
and created a large 96 pixel size BDF font that could work for this project: https://gist.github.com/kmatch98/a53a3241b7be13e3cdc9cae39185713f

jepler:
It's quite likely I made a mistake in my testing.  I'll test again.  Thanks for your patience.

Maybe I tested with the wrong code altogether -- I'm sure the behavior _CHANGED_ depending which file I copied, so that points to me testing with the released version and with some unrelated version with other bugs....

jepler:
I tested again and here's what I found:
* between adafruit-circuitpython-bundle-6.x-mpy-20200811 and the tip of the main (m-----) branch, the vertical position of the BLACK LIVES MATTER text changed.  This text is positioned with x and y and does not use anchored position.  It appears to move UP about a half a line.
 * The tip of the main branch has the shifting left problem
 * This PR's code does not, the problem appears to be fixed.

The first point seems to an incompatible behavior between the last release 2.8.0 on July 24, which may or may not be intentional.  By using anchored_position, I can write code that works with both versions.

FoamyGuy:
Since we've got confirmation that this resolved the issue and an approving review I am going to go ahead and merge this so that hopefully we will minimize the number of folks that run across the bug in their projects. 

With regards to the x,y position changes when not using `anchor_point` and `anchored_position`: I will plan to look into this, I do not think any recent changes intentionally changed the behavior of x/y positioning when `anchored_position` is not in use. So if we can identify when/where it happened we may want to revert it in order to get back to the previous (and documented in many places) behavior. 

It is perhaps worth noting the default positioning for label does put the y position at the center of the label vertically. So if you set a label like:
```
label.x = 0
label.y = 0
```
The expected behavior is that top half of the text will be off the screen. You will only see the bottom half of the text shown in the top left corner of the display.  If the y position getting used moved by half the height my guess is it's related to this default positioning somehow.

PR #88, opened 2020-08-19
gracefully recover if load_glpyhs does not exist on font we are using

### LINKED ISSUE #87: builtin font does not have load_glyphs always
in #85 we add a call to `load_glyphs` inside of `_update_text` and this made the label building  / rendering work faster. 

It looks like in some cases however this is resulting in an error when used with builtin font: 
```
  File "code.py", line 15, in <module>
  File "adafruit_clue.py", line 913, in simple_text_display
  File "adafruit_clue.py", line 117, in __init__
  File "adafruit_clue.py", line 128, in add_text_line
  File "adafruit_display_text/label.py", line 125, in __init__
  File "adafruit_display_text/label.py", line 240, in _update_text
AttributeError: 'BuiltinFont' object has no attribute 'load_glyphs'
```

This was pointed out by a user in the #help-with-circuitpython channel on discord. They were working on the Clue Temperature and Humidity example project from here: https://learn.adafruit.com/adafruit-clue/clue-temperature-and-humidity-monitor I was able to recreate it with that project on a CLUE.

But I do suspect this issue might be more wide spread, perhaps any label that is using terminalio.FONT.

I think we could resolve this by putting the `load_glyphs` call inside of an if statement checking to make sure we aren't currently using `terminalio.FONT`

FoamyGuy:
This resolves #87.

Instead of checking if the font was `terminalio.FONT` I opted for `try/except` and ignoring the AttributeError that results from `load_glyphs` not existing. This lets us avoid having to `import terminalio` and does still solve the crash.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -237,7 +237,13 @@ class Label(displayio.Group):
         else:
             i = 0
         tilegrid_count = i
-        self._font.load_glyphs(new_text + "M")
+
+        try:
+            self._font.load_glyphs(new_text + "M")
+        except AttributeError:
+            # ignore if font does not have load_glyphs
+            pass
+
         y_offset = int(
             (
                 self._font.get_glyph(ord("M")).height


kmatch98:
I tested this and verified that it correctly handles the BuiltinFont `terminalio.FONT` and prevents the observed error.

Thanks for catching this @foamyguy!

PR #92, opened 2020-09-05
Correct documentation errors related to mutability

kmatch98:
Here are some documentation cleanups that I missed in the previous release.  This deletes the sections about immutability and adds more details to the description about when `save_text` is set to `False`.


FoamyGuy (approved):
Looks good to me. Thank you for updating these docs.

PR #93, opened 2020-10-18
add init

eteq:
This is about as trivial as a PR comes (adding a blank file...), but it seemed to me this was necessary to get `python setup.py install` to work correctly to let this by useable on blinka/non-cirtcuitpython?


FoamyGuy (changes requested):
On a raspberry pi 3B+ I was able to install this library successfully with these commands
``` shell
pip3 install adafruit-circuitpython-display-text
```
These also were working successfully for me: 
``` shell
git clone https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git
cd Adafruit_CircuitPython_Display_Text
sudo python3 setup.py install
```
In both cases I tested the install with a Blinka Displayio simpletest script for ili9341 screen. 

Even though it does seem to be working for me as-is, I do think we probably want to include an `__init__.py` file like you've added so that it can be a proper python package. 

Looking over at this library as an example: https://github.com/adafruit/Adafruit_CircuitPython_MCP230xx/blob/master/adafruit_mcp230xx/__init__.py

I see that the empty __init__ file does include a few license lines. 

Can you add these to the file you created and make a new push please? 
```
# SPDX-FileCopyrightText: 2019 Scott Shawcroft and Kevin Matocha for Adafruit Industries
#
# SPDX-License-Identifier: Unlicense
```

tannewt:
I don't think we need a license for an empty file. Nothing to license. :-)

FoamyGuy (approved):
Seems good to me now then. I did a quick test with the anchor point example with both label and  bitmap_label and had no issues. 

Thank you @eteq

Issue #95, opened 2020-11-06
Error in examples/display_text_simpletest.py


lastcoolnameleft:
I have the recent Adabox016 (Adafruit Matrix Portal) and when I try to use the simpletest.py, I get:
```
Traceback (most recent call last):
  File "code.py", line 10, in <module>
AttributeError: 'module' object has no attribute 'DISPLAY'
```

I'm using the library from adafruit-circuitpython-bundle-6.x-mpy-20201030

lastcoolnameleft:
I'm experiencing this issue with `adafruit-circuitpython-matrixportal_m4-en_US-6.0.0-rc.1.uf2`

FoamyGuy:
The example code is assuming that it will get run on a device that has a display built in (in which case it would be available as `board.DISPLAY` within the code). 

But the Matrix Portal needs to do a bit of setup first before it can start using the display. 

There is a [MatrixPortal library](https://github.com/adafruit/Adafruit_CircuitPython_MatrixPortal) which will do lots of the set up for you. If you use that the simpletest example code can be adapted to work on the Matrix portal like this:
```python
import terminalio
from adafruit_display_text import label
from adafruit_matrixportal.matrix import Matrix
matrix = Matrix()
display = matrix.display

text = "Hello\nworld"
text_area = label.Label(terminalio.FONT, text=text)
text_area.x = 1
text_area.y = 4
display.show(text_area)
while True:
    pass
```

FoamyGuy:
@lastcoolnameleft This example has been added in the examples directory now: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/master/examples/display_text_matrixportal.py

I'm going to close this issue for now. But if you end up having trouble when you try this feel free to ping me here or in the discord I can help you out.

lastcoolnameleft:
Thanks for the update and clarification about how/when to use it.  The documentation just pointed me to the repo so I was unaware that the code was for a specific board.  I'm good with it being closed.

PR #96, opened 2020-11-06
Add matrixportal example

FoamyGuy:
Lots of folks have a Matrix Portal now and are possibly being introduced to `displayio` for the first time with it. See #95 for one example. But I think it's popped up a few other places as well. 

The majority of `displayio` examples are written using `board.DISPLAY` assuming that they will be run on a device with a built-in display. Any written this way won't work as is on a Matrix Portal.

This PR adds a copy of the simpletest example that has been modified to work on Matrix Portal.

I do wonder if we want to keep the examples seperated like in this PR, or combined into one file with comment / uncomment a section? Or perhaps there is some other possible technical solution that can work with both types of devices without the user having to worry about it as much?


makermelissa (approved):
Looks good to me.

PR #99, opened 2020-11-17
Wrapping helper function

FoamyGuy:
I think this helper function was originally the PyPortal library.

If we move it to here we can make use of it in all of the various different portal libraries as well as directly from user code. This will make it easier to wrap text for different sized displays.

### DIFF
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -1,6 +1,8 @@
 """
 Display Text module helper functions
 """
+
+
 # The MIT License (MIT)
 #
 # Copyright (c) 2020 Tim C for Adafruit Industries LLC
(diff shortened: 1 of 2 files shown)


FoamyGuy:
@kmatch98 raised a great point in discord: One thing that needs to be tested and hasn't yet is how this code behaves when the supplied text has few enough spaces that it can't break cleanly without splitting a word.

FoamyGuy:
I tried it out and it turns out that it doesn't wrap if it doesn't find a space. I removed the spaces from the first line of text in the wrap example: 
image

It should probably have some kind of fail over to use a hyphen when it can't break at a space. As is it could allow the text to run off the screen. Though it is probably somewhat rare to find text this long in the wild with no space

FoamyGuy:
The new commit addresses this by breaking the long word into chunks and combining them with hyphens as needed.

Here is how the previous example looks now: 
image

ladyada:
nice! @makermelissa may be helpful to factor this out of portal libs

makermelissa (approved):
Tested with my refactored PortalBase library. It adds hyphens as noted. I was going to request a parameter for specifying whether to add them or not, but after thinking about it, I think it's fine.

PR #102, opened 2020-12-05
label: Use new, optional 'ascent', 'descent' properties

jepler:
Each time new glyphs have to be loaded from a font, it can take a long time (hundreds of milliseconds).  Besides which, not all fonts have the "M" character that is used for estimating the ascent.  Instead, use the ascent and descent properties of Font objects, if available. (they are available in bdf and pcf fonts since https://github.com/adafruit/Adafruit_CircuitPython_Bitmap_Font/releases/tag/1.3.0 but terminalio.FONT still doesn't have it)

In particular, before this change the "forkawesome" fonts can't be used, because they have only glyphs starting at Unicode 0xf000 (called the Private Use Area), and may affect webdings/wingdings type fonts as well. (forkawesome fonts converted to pcf are [on my blog](https://emergent.unpythonic.net/01606790241))

This changes the layout of text slightly.  For instance, the height of the "M" glyph in GothamBlack-50.bdf is 35, but the Ascent of the font is 40 (and some characters, such as Å, are taller than the ascent at 45 pixels high).  So in this case, text may be shifted down by about 2 to 3 pixels (half of the difference between 35 and 40, as fonts are positioned vertically relative to the half ascent).

To test it, grab https://media.unpythonic.net/emergent-files/01606790241/forkawesome-36.pcf and run the following code (I used a pyportal titano):
```python
from displayio import Group

font = load_font("/forkawesome-36.pcf")
w,h,dx,dy = font.get_bounding_box()

glyphs = "".join(chr(0xf000 + i) for i in range(8))

group = Group()
label = Label(font=font, text=glyphs, background_color=0x111111)
label.anchor_point = (0, 0)
label.anchored_position = (0,0)

group.append(label)
board.DISPLAY.show(group)
while True:
    pass
```

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -147,28 +147,17 @@ class Label(displayio.Group):
             y_box_offset = self._boundingbox[1]
 
         else:  # draw a "loose" bounding box to include any ascenders/descenders.
-
-            # check a few glyphs for maximum ascender and descender height
-            # Enhancement: it would be preferred to access the font "FONT_ASCENT" and
-            # "FONT_DESCENT" in the imported BDF file
-            glyphs = "M j'"  # choose glyphs with highest ascender and lowest
-            # descender, will depend upon font used
-            ascender_max = descender_max = 0
-            for char in glyphs:
-                this_glyph = self._font.get_glyph(ord(char))
-                if this_glyph:
-                    ascender_max = max(ascender_max, this_glyph.height + this_glyph.dy)
-                    descender_max = max(descender_max, -this_glyph.dy)
+            ascent, descent = self._get_ascent_descent()
 
             box_width = self._boundingbox[2] + self._padding_left + self._padding_right
             x_box_offset = -self._padding_left
             box_height = (
-                (ascender_max + descender_max)
+                (ascent + descent)
                 + int((lines - 1) * self.height * self._line_spacing)
                 + self._padding_top
                 + self._padding_bottom
             )
-            y_box_offset = -ascender_max + y_offset - self._padding_top
+            y_box_offset = -ascent + y_offset - self._padding_top
 
         box_width = max(0, box_width)  # remove any negative values
         box_height = max(0, box_height)  # remove any negative values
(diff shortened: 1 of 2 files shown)


adafruit_display_text/label.py
```diff
@@ -248,13 +256,7 @@ def _update_text(
             i = 0
         tilegrid_count = i
 
-        try:
```
FoamyGuy, line 251: I think we may need to bring back this try / except. 

Attempting to use `terminalio.FONT` (simpletest script) I am getting this error with this version from this PR:
```
  File "code.py", line 7, in <module>
  File "/lib/adafruit_display_text/label.py", line 135, in __init__
  File "/lib/adafruit_display_text/label.py", line 259, in _update_text
  File "/lib/adafruit_display_text/label.py", line 192, in _get_ascent
  File "/lib/adafruit_display_text/label.py", line 181, in _get_ascent_descent
AttributeError: 'BuiltinFont' object has no attribute 'load_glyphs'
```

FoamyGuy (changes requested):
I tested with the forkawesome font and code provided on a PyPortal and that is working well.

Once we handle the one issue with BuiltinFont I think this is good to go.

FoamyGuy:
I will also follow up with @kmatch98 after this gets merged and we can apply similar changes in the BitmapLabel.

adafruit_display_text/label.py
FoamyGuy: I had the right idea in needing to catch an attribute error, but wrong place.  Up inside of `_get_ascent_descent` was the correct place to catch and ignore this error since the built-in font doesn't use that.

FoamyGuy (approved):
I added a few commits, one that falls back safely for the builtin font. This resolves the issue with simpletest, and any other labels using the builtin font.

One that added extra spaces in a few docstrings inside of `bitmap_label` these lines were otherwise untouched by the PR, but I think it came up due to an update in Sphinx version. Had to update my local tool to replicate and fix.

I tested the current version successfully with the test scripts from this PR and the simpletest in the examples directory

FoamyGuy:
I would like to get another librarian to take one more quick look at this if possible since I made the final commits.

My full test code with imports and all is:
```python
import board
from displayio import Group

font = bitmap_font.load_font("/fonts/forkawesome-36.pcf")
w,h,dx,dy = font.get_bounding_box()

glyphs = "".join(chr(0xf000 + i) for i in range(8))

group = Group()
label = label.Label(font=font, text=glyphs, background_color=0x111111)
label.anchor_point = (0, 0)
label.anchored_position = (0,0)

group.append(label)
board.DISPLAY.show(group)
while True:
    pass
```
This is ready to merge after one more look-over. Thanks in advance!

tannewt (approved):
Code looks fine to me. I didn't test it.

PR #105, opened 2020-12-17
Fixed wrapping function when text length = maxchars

### LINKED ISSUE #104: str index out of range for wrap_text_to_lines
I'm getting IndexError: str index out of range on wrapping strings with:
```
from adafruit_display_text import wrap_text_to_lines
print(wrap_text_to_lines("Hello World!", 5))
```

makermelissa:
Fixes #104. This now works on short text when the max_chars is equal to the text words.

### DIFF
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -53,6 +53,7 @@ def wrap_text_to_lines(string, max_chars):
             for part in chunks(w, max_chars - 1):
                 parts.append("{}-".format(part))
             the_lines.extend(parts[:-1])
+            # Remove the last hyphen on the last chunk
             the_line = parts[-1][:-1]
             continue
 
@@ -63,6 +64,9 @@ def wrap_text_to_lines(string, max_chars):
             the_line = "" + w
     if the_line:  # Last line remaining
         the_lines.append(the_line)
+    # Remove any blank lines
+    while not the_lines[0]:
+        del the_lines[0]
     # Remove first space from first line:
     if the_lines[0][0] == " ":
         the_lines[0] = the_lines[0][1:]
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -49,6 +49,8 @@ def wrap_text_to_lines(string, max_chars):
     the_line = ""
     for w in words:
         if len(w) > max_chars:
+            if the_line:  # add what we had stored
+                the_lines.append(the_line)
             parts = []
             for part in chunks(w, max_chars - 1):
                 parts.append("{}-".format(part))
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -53,12 +53,13 @@ def wrap_text_to_lines(string, max_chars):
             for part in chunks(w, max_chars - 1):
                 parts.append("{}-".format(part))
             the_lines.extend(parts[:-1])
-            # Remove the last hyphen on the last chunk
             the_line = parts[-1][:-1]
             continue
 
         if len(the_line + " " + w) <= max_chars:
             the_line += " " + w
+        elif not the_line and len(w) == max_chars:
+            the_lines.append(w)
         else:
             the_lines.append(the_line)
             the_line = "" + w


FoamyGuy:
Thanks for this fix, I will test this out tonight.

FoamyGuy:
I think we still have some kind of issue here. This line no longer throws an exception but it's also dropping the entire first word:
```
>>> print(wrap_text_to_lines("Hello World!", 5))
['Worl-', 'd!']
```

adafruit_display_text/__init__.py
```diff
@@ -63,6 +64,9 @@ def chunks(lst, n):
             the_line = "" + w
```
FoamyGuy, line 64: I think the root of the issue is this line needs to be swapped with the line above it. This should come first. It's setting `the_line` variable, and we want to add it to the list after it gets set.

These being in this order is why it previously was adding `None` to the list which caused the exception. 

I think if we swap these two lines then we don't need the check to remove blank ones any more.

FoamyGuy: I tested with this change and get correct output for "Hello World!":
```
>>> print(wrap_text_to_lines("Hello World!", 5))
['Hello', 'Worl-', 'd!']

```

makermelissa: Ok cool. Make sure it still works on the long text. I think I had swapped it around briefly and it worked in one case but not the other. I thought I had tested it working on both when I submitted this PR, but apparently not. :)

FoamyGuy: Ah, you are right this does cause problems for longer sentences. Using the text in the wraptest example it is losing many of the words:
```
>>> text = (
...     "Lorem ipsum dolor sit amet, consectetur adipiscing elit, "
...     "sed do eiusmod tempor incididunt ut labore et dolore magna "
...     "aliqua. Ut enim ad minim veniam, quis nostrud exercitation "
...     "ullamco laboris nisi ut aliquip ex ea commodo consequat."
... )
>>> wrap_text_to_lines(text, 28)
['consectetur', 'sed', 'incididunt', 'dolore', 'ad', 'nostrud', 'laboris', 'ea', 'ea commodo consequat.']
```

Maybe these two lines have to stay in the original order, but a new branch added to the logic check for words that are the exact same length as the wrap max per line amount.

I will tinker with it some more tonight.

makermelissa: I tried out this change and `print(wrap_text_to_lines("Hello World", 5))` outputs:
```
['Hello', 'World', 'World']
```
So it seems there is more to this.

makermelissa:
Ok, I made a small adjustment. It seems the Hello World was more of an edge case and I adjusted for that.

adafruit_display_text/__init__.py
FoamyGuy: I think i'm getting the correct output from the current code from this PR. 
```python
>>> print(wrap_text_to_lines("Hello World", 5))
['Hello', 'World']
```
I will be back a bit later to do some more testing

makermelissa: I tested it with:
 * "Hello World"
 * "Hello4 World"
 * "Hello World!"
 * The text from the simple test
 and they all seemed to work good. Please feel free to try it out with some other things.

FoamyGuy: The majority of everything I tested worked well. I did find one more example with some odd behavior:
```
>>> print(wrap_text_to_lines("The quick brown fox jumped", 5))
['The', 'quick', 'brown', 'jump-', 'ed']
```
Somehow it's losing "fox" from this input.

FoamyGuy: Okay I think I figured that out. Whenever it finds a word bigger than the line limit and breaks it down with hyphens it's "throwing away" the current value of `the_line`. To fix this we can check for `the_line` and append it:
```
if the_line:  # add what we had stored
    the_lines.append(the_line)
```

Inside the hyphen breakdown check here:
https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/2a786af5e49be273b14a3b6846f452711bd44242/adafruit_display_text/__init__.py#L51

makermelissa: Thanks @foamyguy, I'll give that a try and if the results look good, I'll submit the change.

makermelissa: Ok, that seemed to work. Here's my test script at this point:
```
from adafruit_display_text import wrap_text_to_lines

text = (
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit, "
    "sed do eiusmod tempor incididunt ut labore et dolore magna "
    "aliqua. Ut enim ad minim veniam, quis nostrud exercitation "
    "ullamco laboris nisi ut aliquip ex ea commodo consequat."
)

text2 = (
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit, "
    "sed do eiusmod tempor incididunt ut labore et dolore magna "
    "aliqua. Ut enim ad minim veniam, quis nostrud exercitation "
    "ullamco laboris nisi ut aliquip ex ea commodo consequat."
)

print(wrap_text_to_lines("Hello World", 5))
print(wrap_text_to_lines("Hello! World", 5))
print(wrap_text_to_lines("Hello World!", 5))
print(wrap_text_to_lines(text, 28))
print(wrap_text_to_lines("The quick brown fox jumped over the lazy dog", 5))
```

FoamyGuy (approved):
Looks good to me. Thank you!

makermelissa:
Thanks @FoamyGuy

PR #106, opened 2021-01-14
Ran pre-commit, added licenses

dherrada:
@ladyada Also not sure which license to give the font here


FoamyGuy:
I will make a PR to switch to a font file with a specific licence. There are some permissively licenced fonts available here: https://www.theleagueofmoveabletype.com/

There is an example of the "Open Font License" they are released under in this learn guide project: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/master/MagTag_Progress_Displays/fonts/leaguespartan.license

PR #107, opened 2021-01-17
adding font with license file

FoamyGuy:
Font change as mentioned in #106

Also fix one of the examples for built-in screens.


ladyada (approved):

FoamyGuy:
I think this may not be able to pass without the changes from #106 

The easiest way that I know of to get everything back together is merge from the branch used for #106 and this PR can supersede that one, it will all of those commits, plus mine from here. I'll give that a try but open to other ideas if there is an easier way that I am missing.

PR #111, opened 2021-01-31
Adding the wrap_by_pixels function

FoamyGuy:
Jeff wrote this nice helper function for wrapping text to a number of pixels instead of characters.

This PR adds that function as well as an example that illustrates it's usage.

One quirk worth consideration is that the new wrap_by_pixels function returns a string with newlines inserted. Whereas the older wrap_by_characters function returns a list of the lines, and joining with the newline is left up to the code using it. I think I like the way this one does it better with the string. But the other one came from the PyPortal libraries originally, perhaps there is something in them that is using the list. Maybe it would be worthwhile to add a parameter to both that controls this behavior?

### DIFF
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -41,10 +41,7 @@ def wrap_text_to_pixels(string, max_width, font=None, indent0="", indent1=""):
     swidth = measure(" ")
     firstword = True
     for line_in_input in string.split("\n"):
-
-        print("line: {}".format(line_in_input))
         for index, word in enumerate(line_in_input.split(" ")):
-            print("word: {}".format(word))
             wwidth = measure(word)
             word_parts = []
             cur_part = ""
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -40,41 +40,52 @@ def wrap_text_to_pixels(string, max_width, font=None, indent0="", indent1=""):
     width = measure(indent0)
     swidth = measure(" ")
     firstword = True
-    for word in string.split():
-        wwidth = measure(word)
-        word_parts = []
-        cur_part = ""
-        if wwidth > max_width:
-            if partial:
-                lines.append("".join(partial))
-                partial = []
-            for char in word:
-                if measure(cur_part) + measure(char) + measure("-") > max_width:
-                    word_parts.append(cur_part + "-")
-                    cur_part = char
-                else:
-                    cur_part += char
-            if cur_part:
-                word_parts.append(cur_part)
-            for line in word_parts[:-1]:
-                lines.append(line)
-            partial.append(word_parts[-1])
-            width = measure(word_parts[-1])
-            if firstword:
-                firstword = False
-        else:
-            if firstword:
-                partial.append(word)
-                firstword = False
-                width += wwidth
-            elif width + swidth + wwidth < max_width:
-                partial.append(" ")
-                partial.append(word)
-                width += wwidth + swidth
+    for line_in_input in string.split("\n"):
+
+        print("line: {}".format(line_in_input))
+        for index, word in enumerate(line_in_input.split(" ")):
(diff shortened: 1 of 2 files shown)


ladyada:
id kinda prefer the list-return version - can you adapt it to match?

dherrada:
@FoamyGuy I'll test this sometime in the next few days

kmatch98:
I ran several scenarios to sort out the behavior with miscellaneous white space.  I found one bug that adds a leading newline.  The other two things may be your planned behavior, so my only suggestion is to add a couple lines of documentation to clarify how newlines and leading/trailing whitespace is treated.

#### Bug?
Input string: `text = "THISISasuperlongstringwithnospacesthatwillnotfit"`
Output: `['', 'THISISasuperlongstring-', 'withnospacesthatwillno-', 'tfit']`
Adds an extra newline at the beginning.

#### For Documentation
Input string: `text = "\t A short \ntest"`
Output: `['A short test']`
Behavior: Newlines and tabs are obliterated. 
_(Unrelated to this, I found that `label` and `bitmap_label` both ignore tabs.  I'll add an issue.)_

#### For Documentation
Input string: `text = " Leading and trailing spaces are deleted.    "`
Output: `['Leading and trailing', 'spaces are deleted.']`
Behavior: Leading and trailing spaces are deleted.

FoamyGuy:
Thanks for looking into this @kmatch98! 

Definitely a bug on the first one.

I lean toward bug on `\n`s getting  removed as well. It's probably best not to use them if you want the function to do the wrapping, but if the user includes them we should try to keep them if we can.  Tabs we can look at once they get fixed in label and bitmap_label. 

Agreed about documenting the removal of leading and trailing spaces. For leading spaces the `indend0` parameter can be used to add some spaces ( or anything) to the beginning

I'll work on making these changes to fix and document the above.

dherrada:
@FoamyGuy Currently pretty swamped with other stuff so I'm not sure when I'm gonna get to this. If you're fine waiting a bit (looking like a week at least) I'd still be down to review this, but otherwise I won't be offended if someone else reviews it

FoamyGuy:
Okay,  I think this is ready for review again. The latest commits resolve the issues noted above in the following ways:

- **Very long word results in extra newline at the beginning.** This was due to the longword logic always starting on the next line down instead of picking up in the current spot of the line. That coupled with the `indent0` value was causing the extra line at the beginning. Newest version just starts right where it's at in the partial line instead of jumping down to the next. This avoids the extra newline in the single long word scenario. 

- **tabs and newlines removed.**  Tabs we will need to deal with inside of label and bitmap label. So I didn't do anything specific here for them yet. Newlines no longer get removed, and are now properly respected. So the user can have newlines in the input string to get extra empty lines in the output.

- **leading and trailing whitespace removed**. This behavior is now documented along with a note about the `indent0` and `indent1` parameters that can be used to control leading whitespace if desired.

kmatch98:
I tested and the this fixes the issues mentioned above.  I also verified that the leading and trailing spaces are now retained.  

I checked the operation of the `indent0` and `indent1`.  I noticed that using the `indent1` option always leaves an extra line at the end.  Not a big deal, but it's unexpected.

```python

text = "CircuitPython is a programming language"
print(wrap_text_to_pixels(text, WRAP_WIDTH, font, indent0="*", indent1="-"))
```
Output:
```
['*CircuitPython is a', '-programming language', '-']
```

Thanks for these fixes!

FoamyGuy:
Nice catch @kmatch98, thank you! `indent1` was getting added an extra time by some code adding `partial` pat the end an extra time.

The latest commit resolves it, the returned list no longer has an extra item in it.

kmatch98:
I verified this update, it definitely solves the problem.  Thanks for getting this merged into the library and fixing all these details!

Looks good to me.

dherrada (approved):
Looks good to me. Tested on a PyPortal Titano

Issue #103, opened 2020-12-15
Should setting background_tight=True change the label's bounding box?


endico:
I have two labels with different sized fonts which I would like to place next to each other along the same baseline. My solution was to place the one on the right in relation to the one on the left using the information in the left's bounding box. This is still a bit off though because the labels have different font sizes so each label has a different amount of padding at the bottom (where the descenders go).

These labels have a contrasting background color and I noticed that when I set background_tight on a label containing numerals that the padding at the bottom disappears and the bottom of the label seems to coincide with the font's baseline. Unfortunately, doing this doesn't change the bounding box. If it did, that would make laying out my numerals a lot easier. And the code would still work after changing the font, adjusting margins or possibly dragging multiple labels around with a touchscreen.

FoamyGuy helped me with a workaround that works fine in my app but this seems useful. I guess the downside is that it won't be very useful for text since a lot of it will have descenders. 

Maybe its not worth the trouble? I guess this is more hacky than I originally thought and maybe it would break existing  code. However, it does seem to make sense that the bounding box should move when the padding changes.

endico:
I found a property of adafruit_bitmap_font.bdf called ascent which is: The number of pixels above the baseline of a typical ascender. The ascender property solves my layout problem nicely although when I try to print font.ascent i get "AttributeError: 'BDF' object has no attribute 'ascent'"
https://circuitpython.readthedocs.io/projects/bitmap-font/en/latest/api.html#module-adafruit_bitmap_font.bdf

endico:
`import board
from adafruit_display_text import label
import displayio
from adafruit_bitmap_font import bitmap_font

display = board.DISPLAY

main_group = displayio.Group()
medium_font = bitmap_font.load_font("/fonts/Exo-SemiBold-18.bdf")
big_font = bitmap_font.load_font("/fonts/Exo-Bold-42.bdf")
# print("big_font.ascent= ", big_font.ascent)

margin = 10

left_text = label.Label(
    big_font,
    max_glyphs=6,
    text="0",
    color=0x000000,
#    background_tight=True,
    background_color=0x999999,
    anchor_point=(0.0, 0.0),
    anchored_position=(margin, margin),
)
main_group.append(left_text)

right_text = label.Label(
    medium_font,
    text="MG",
    color=0x000000,
    background_color=0x999999,
    x=0,
    y=0,
)
# Pretend to assign new value from json
#
left_text.text = "32.47"

[...]

Issue #100, opened 2020-11-19
Feature request: anchor relative to baseline


PaintYourDragon:
Anchoring text along a baseline (rather than or in addition to bounding rect) would be very Very _very_ helpful. Maybe this could be indicated without breaking existing code by expressing a Label’s anchor_point as a single value or single-element tuple rather than a 2-element tuple (so it still can do left/center/right alignment, while y coord is then just the baseline, period).

As it currently stands, text bounds can vary with descenders, punctuation, etc. This makes it impossible to guarantee a consistent baseline when there are multiple Label elements that are trying to align as a well-formatted and legible unit. Also, for text that changes value, the current use of bounding rect alignment can cause Label elements to move up or down as the value changes, rather than maintaining their original baseline.

Attached image shows example of desired baseline behavior, which was approximated with some hacks. For the first two items on each line, baseline alignment is forced by using bottom alignment and applying upper() to the location strings (though none of these locations have descenders…but if any of them did, say a “j” or “g,” and were not upper() filtered, those descenders would alter the text bounds, placing the Label bottom-aligned by bounds, but with a baseline that jumps up a few pixels and Looks Bad). The third element on each line here is also bottom aligned…but with +2 added to each anchored position’s Y value…it works here because we know that a comma in this font and this size descends 2 pixels below the baseline, but that’s just a quick workaround and not going to be true for all fonts and sizes. Baseline anchoring would take care of all of that.

![baseline](https://user-images.githubusercontent.com/887611/99628102-54662b00-29ea-11eb-9cac-081f4ba3faaf.jpg)

kmatch98:
Fully agreed that the baseline is a major consternation but was based on the (historical?) version of `label`. It uses the height of an ‘M’ to define the baseline using this code (line 199):

```
y_offset = int((self._font.get_glyph(ord("M")).height) / 2)
```

The baseline position is actually defined in the BDF font specification but is currently ignored in the current code. 

On a related note, the ascender/descender height should be included in the BDF file but is not currently utilized. Instead, a few characters are checked to test the ascender/descender height. These dimensions is used for the background sizing along with tight_padding. 

Any changes to the baseline should be propagated to label and bitmap_label for consistency. 

One complexity with baseline positioning is that two ways are currently available to set the position, either `(x,y)` or by `anchor_position`.  

If we can think of a way to consolidate this perhaps this would be more well-defined (but maybe more confusing?). 

One proposal: 
[...]

PR #115, opened 2021-02-07
New parameter and example creation for testing and future reference

jposada202020:
With this change, we are adding a new parameter called base_alignment in the label library.  Also we are including some tests to compare old behavior vs New behavior. This will address issues #103 and #100. 
For back-compatibility parameter is set to False by default, that way old code created will still function with the new version

### DIFF
--- a/examples/display_text_label_align_baseline_comparison.py
+++ b/examples/display_text_label_align_baseline_comparison.py
@@ -1,0 +1,71 @@
+import board
+from adafruit_display_text import label
+from adafruit_bitmap_font import bitmap_font
+import displayio
+import time
+
+display = board.DISPLAY
+
+# Font definition. You can choose any two fonts available in your system
+MEDIUM_FONT = bitmap_font.load_font("Helvetica-Bold-16.bdf")
+BIG_FONT = bitmap_font.load_font("LeagueSpartan-Bold-16.bdf")
+
+# Test parameters
+TEXT_BACKGROUND_COLOR = 0x990099
+TEXT_COLOR = 0x00FF00
+X_COL_POS = [10, 90]  # X position of the Label boxes
+TEST_TEXT = [
+    "aApPqQ.",
+    "ppppppp",
+    "llllll",
+    "oooooo",
+    "ñúÇèüß",
+]  # Iteration text for testing
+LOCALISATION_Y = [40, 80, 120, 160, 200]  # Y coordinate of the text labels
+NUMBER_TEXT_LABELS = len(LOCALISATION_Y) * len(
+    X_COL_POS
+)  # Calculation for Display.Group function
+TIME_BETWEEN_TEXTS = 5  # Seconds
+BASE_ALIGNMENT_OPTIONS = [False, True]
+
+# Test
+for behaviour in BASE_ALIGNMENT_OPTIONS:
+    for text_test in TEST_TEXT:
+        main_group = displayio.Group(max_size=NUMBER_TEXT_LABELS + 1)
+        for position in LOCALISATION_Y:
+            text = label.Label(
+                font=MEDIUM_FONT,
+                text=text_test,
+                color=TEXT_COLOR,
+                background_tight=True,
+                background_color=TEXT_BACKGROUND_COLOR,
+                x=X_COL_POS[0],
+                y=position,
+                base_alignment=behaviour,
+            )
+            main_group.append(text)
+            text = label.Label(
+                font=BIG_FONT,
+                text=text_test,
+                color=TEXT_COLOR,
+                background_tight=False,
+                background_color=TEXT_BACKGROUND_COLOR,
+                x=X_COL_POS[1],
+                y=position,
+                base_alignment=behaviour,
+            )
+            main_group.append(text)
+            display.show(main_group)
+            display.refresh()
+        time.sleep(TIME_BETWEEN_TEXTS)
+
+
+bitmap = displayio.Bitmap(280, 5, 2)
+palette = displayio.Palette(2)
+palette[0] = 0x004400
+palette[1] = 0x00FFFF
+tile_grid2 = displayio.TileGrid(bitmap, pixel_shader=palette, x=10, y=160)
+main_group.append(tile_grid2)
+
+while True:
+    pass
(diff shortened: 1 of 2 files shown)


jposada202020:
@FoamyGuy 
```
  import displayio
  from blinka_displayio_pygamedisplay import PyGameDisplay
  import terminalio
  import time
  import random
  
  display = PyGameDisplay(width=320, height=240)
  
  # Create the text label
  main_group = displayio.Group(max_size=18)
  
  # Test parameters
  TEXT_BACKGROUND_COLOR = [0xe0433, 0x990099, 0xd6d714, 0xd73014, 0x1415d7, 0x71646f]
  FONT_TESTS = [bitmap_font.load_font("Fayette-HandwrittenScript-24.bdf"),
                bitmap_font.load_font("LeagueSpartan-Bold-16.bdf"),
                terminalio.FONT]
  TEXT_COLOR = 0x00FF00
  X_COL_POS = [10, 110]  # X position of the Label boxes
  TEST_TEXT = ["aApPqQ.", 'ñúÇèüß']  # Iteration text for testing
  LOCALISATION_Y = [40, 80]  # Y coordinate of the text labels
  NUMBER_TEXT_LABELS = len(LOCALISATION_Y) * len(X_COL_POS)  # Calculation for Display.Group function
  TIME_BETWEEN_TEXTS = 3  # Seconds
  TEST_PADDING_VALUES = [-3, 0, 15]
  TEST_CHOICES = [True, False]
  BASE_ALIGNMENT_OPTIONS = [False, True]
  
  
  for behaviour in BASE_ALIGNMENT_OPTIONS:
      for padding in TEST_PADDING_VALUES:
          for text_test in TEST_TEXT:
              main_group = displayio.Group(max_size=NUMBER_TEXT_LABELS + 1)
              for position in LOCALISATION_Y:
                  text = label.Label(
                      font=FONT_TESTS[0],
                      text=text_test,
                      color=TEXT_COLOR,
                      background_tight=random.choice(TEST_CHOICES),
                      background_color=TEXT_BACKGROUND_COLOR[0],
                      x=X_COL_POS[0],
                      y=position,
                      base_alignment=behaviour,
                  )
                  main_group.append(text)
                  text = label.Label(
                      font=FONT_TESTS[1],
                      text=text_test,
                      color=TEXT_COLOR,
                      background_tight=False,
                      background_color=TEXT_BACKGROUND_COLOR[1],
                      x=X_COL_POS[1],
                      y=position,
                      base_alignment=behaviour,
                  )
                  main_group.append(text)
                  display.show(main_group)
                  display.refresh()
              time.sleep(TIME_BETWEEN_TEXTS)

```

examples/display_text_label_align_baseline_comparison.py
```diff
@@ -0,0 +1,79 @@
+# SPDX-FileCopyrightText: 2021 ladyada for Adafruit Industries
```
FoamyGuy, line 1: You can change the copyright to your name in this example script that you wrote if you are comfortable with that. 

I think we should either add one extra example that is much simpler. or else modify this one to be simpler. I'll leave it up to you. I can see the upside of keeping this one for more comprehensive test.

I think we should provide something like `display_text_label_align_baseline_simpletest.py` that will contain nothing but two labels with different fonts with the new parameter enabled to show how they can easily align with each other.

FoamyGuy (changes requested):
Thanks @jposada202020. I'm not sure what I saw last time but it does seem all good now with the Fayette font.

I think is pretty close to ready. It seems that when I merged the scale PRs it caused merge conflicts here. I think it can be resolved if you merge master into this branch. Let me know if you'd like help though, I'm happy to help out.

I'd like to get a simplified example if we can as well.

After those things I think this is probably good to go! Thanks again for working on this!

jposada202020:
@FoamyGuy Thanks for the revision, thanks for all the suggestions, For the code, you are right, I was just trying to find where the problem was :) ...I think I will use original code by @endico in issue #103 for the example. And also make the changes as suggested. Thanks for the review. And for your time 👍🏼

jposada202020:
...I forgot "it caused merge conflicts here. I think it can be resolved if you merge master into this branch. Let me know if you'd like help though, I'm happy to help out." Yes, I will need help with that. I will ping you when other changes are ready

kmatch98:
I ran the example code with a few different font files and it works great!  

The Large font that's in the example wasn't in the repository `fonts` folder.  Maybe add another large version of the LeagueSpartan to include in the repo to highlight this new feature?

examples/display_text_label_align_baseline_comparison.py
```diff
@@ -0,0 +1,83 @@
...
+
+TEXT_RIGHT = "MG"
+TEXT_LEFT = "32.47"
+
+while True:
+    main_group = displayio.Group()
+    print("Base Alignment parameter False")
+
+    # Create label
+    left_text = label.Label(
```
FoamyGuy, line 29: instead of creating new labels each iteration in the main loop, lets try to make the labels before the main loop and then either only update which one is showing inside the main loop.  Or if you want you could make 4 labels 2 sets one with the new argument, true and one with it false. Similar to how this comparison example is: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/master/examples/display_text_label_vs_bitmap_label_comparison.py . If you need some more specific ideas how to do it let me know.

FoamyGuy (commented):
Thanks for working on this! I see the conflicts are handled now.

jposada202020:
@FoamyGuy I changed the example to include both 4 labels in the screen at all times, Thanks for the comments reviews, and help with this PR.

FoamyGuy (approved):
I tested these changes successfully on PyPortal and MagTag. Changes look good to me. Thanks for working on this @jposada202020!

Here is how the new example looks on a MagTag:
image

I changed the big font to one I had available and adjusted the positions slightly to fit the magtag screen better.

jposada202020:
@FoamyGuy Team Work!!!! Thanks! You save us some time for the documentation :). Thank you

kmatch98:
@jposada202020 this is really helpful. Thanks for adding.  Do you plan to do the same for the bitmap_label version?

jposada202020:
@kmatch98 

Yes no problem. I will like to take a look at bitmap_label at the same time to do the Documentation (Including/Excluding features compared with Label) if that is ok?. So give me until the weekend :)

Issue #118, opened 2021-02-11
Tabs are ignored in `label` and `bitmap_label`


kmatch98:
Currently, all tabs `'\n'` are ignored in both label and bitmap_label.

lesamouraipourpre:
That could be a challenge to define what a tab is. Do you define it as a fixed number of pixels, eg 25px, or do you define it based on the width of a character, eg 8 spaces. As a programmer, I would go with 4 spaces.

Also, how would it behave if the text is not left anchored?

jposada202020:
Agree with lesamouraipourpre, at the end of the day the 4 spaces will get translated to pixels in the library according to the Font used

jposada202020:
@kmatch98 @FoamyGuy if you are not currently wrking at this I can take a look tonight. Let me know. :)

FoamyGuy:
@jposada202020 I'd say go for it. Thank you! I'm going to test out the bitmap_label alignment PR next when I get back to it. This was a bit lower down the list for me, so am happy to have you take it on if you are willing to.

PR #121, opened 2021-02-20
Changing logic to include tab replacement and glyph calculation

jposada202020:
Present changes allows the library to change the "\t" tabs by 4 spaces. The library then will use all the normal logic to take care of these. Present solution was chosen for two main reasons:

1. This could be done by a re expression in the form of 
```
if re.search('\t',  text):
    text = text.replace('\t', '    ')
```
However this will imply import the re library. Not sure that this is what why want.
2. We could use a Post treatment using glyph.shift_x in the _update_text module
```
 if not glyph:
    if character == 't':
        glyph = self._font.get_glyph(ord(" "))
        glyph.shift_x = glyph.shift_x * 4
else:
    continue
```
At the end,  we need to return the parameter to the original value with. This needs to be done as the library uses glyph<s parameters for calculations and display:
```
if character == '\t':
    glyph.shift_x = glyph.shift_x // 4
```
This is a hack to disguise a tab as an space with a different shift_x parameter as "\t" will not be found as a glyph. So less Pythonic.

This was tested with the following code:
```
# SPDX-FileCopyrightText: 2021 Jose David Montoya
# SPDX-License-Identifier: MIT

"""
This example shows the changes to verify the tab.
"""

from blinka_displayio_pygamedisplay import PyGameDisplay
import displayio
import terminalio

display = PyGameDisplay(width=320, height=240)


# Test parameters
TEXT_BACKGROUND_COLOR = [0xe0433, 0x990099, 0xd6d714, 0xd73014, 0x1415d7, 0x71646f]
TEXT_COLOR = 0x00FF00
FONT_TESTS = [bitmap_font.load_font("IBMPlexMono-Medium-24_jep.bdf"),
              bitmap_font.load_font("LeagueSpartan-Bold-16.bdf"),
              terminalio.FONT]

main_group = displayio.Group(max_size=8)
text = ["\tñ    \t Mp", "    ñ\tp     M", "\tp Mñ\t    ", " tab\ttest "]
ypos = [100, 130, 160, 190]
text_testing = list()
[text_testing.append(label.Label(FONT_TESTS[0],
                                 text=text_iter,
                                 background_color=TEXT_BACKGROUND_COLOR[1],
                                 x=40, y=y_iter))
                                 for text_iter, y_iter in zip(text, ypos)]

[main_group.append(_) for _ in text_testing]

display.show(main_group)

while True:
    pass
```

This produce the following result
image

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -91,6 +91,7 @@ class Label(displayio.Group):
         if not max_glyphs and not text:
             raise RuntimeError("Please provide a max size, or initial text")
         if not max_glyphs:
+            text = "    ".join(text.split("\t"))
             max_glyphs = len(text)
         # add one to max_size for the background bitmap tileGrid
 
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -61,7 +61,9 @@ class Label(displayio.Group):
      containing x,y pixel coordinates.
     :param int scale: Integer value of the pixel scaling
     :param bool base_alignment: when True allows to align text label to the baseline.
-     This is helpful when two or more labels need to be aligned to the same baseline"""
+     This is helpful when two or more labels need to be aligned to the same baseline
+    :param tuple(int, str): tuple with tab character replace information. When (4, " ")
+     will indicate a tab replacement of 4 spaces, defaults to 4 spaces by tab character"""
 
     # pylint: disable=too-many-instance-attributes, too-many-locals
     # This has a lot of getters/setters, maybe it needs cleanup.
@@ -86,11 +88,13 @@ class Label(displayio.Group):
         anchored_position=None,
         scale=1,
         base_alignment=False,
+        tab_replacement=(4, " "),
         **kwargs
     ):
         if not max_glyphs and not text:
             raise RuntimeError("Please provide a max size, or initial text")
-        text = "    ".join(text.split("\t"))
+        self.tab_text = tab_replacement[1] * tab_replacement[0]
+        text = self.tab_text.join(text.split("\t"))
         if not max_glyphs:
             max_glyphs = len(text)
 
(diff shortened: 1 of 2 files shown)


jposada202020:
@FoamyGuy @kmatch98 please review the followings changes. This will hopefully address partially issue #118. Thanks for your comments.

adafruit_display_text/label.py
```diff
@@ -91,7 +91,9 @@ def __init__(
         if not max_glyphs and not text:
             raise RuntimeError("Please provide a max size, or initial text")
```
kmatch98, line 92: You could save a line and just put the `text = "    ".join(text.split("\t"))` before the `if not max_glyphs:` statement.

It does this calculation either way, so just do it once and then do the `max_glyphs` check.

kmatch98 (commented):
@jposada202020 this works as expected.

Just one minor suggestion on the code.  Also, if you are interested, will you please apply similar "\t" handling for bitmap_label?

Thanks for adding this in!

adafruit_display_text/label.py
jposada202020: Thanks for the feedback, good point, I will change it and re-do the test just to be 100% sure. Thanks @kmatch98 :)

jposada202020:
@kmatch98 yes I am in, I will do the bitmap_label too. just wanted to be open about the different approaches and let you folks  discuss if needed :).

kmatch98:
Looks good to me.  Thanks for the quick update!

FoamyGuy (approved):
This looks good to me. I tested with Blinka_Displayio and PyPortal. 

I wonder if it will throw someone off that the string they get back might be different from the input string if the access the `text` property. It may be worthwhile eventually to add a parameter to control this. But I'm not sure if the complexity of that would be worth it unless folks are really having a problem with it.

I think we need to wait for the pylint fix to merge maybe so that actions will pass.

jposada202020:
@FoamyGuy Changes are made for the tab parameter. @kmatch98 I changed also for the bitmap_label :). We discuss in the stream after a suggestion from Hugo, that would be good to have that, that way the user can decide how many spaces, dashes or other characters he want the replace the \t for. I was thinking about the setter, but in the same line of thought as Foamyguy, I would like to keep the library simple, and only work if some users ask for that feature as a setter. Thanks.

jposada202020:
PR closed as changes were introduced in  release 2.15.4 - Fix for tab characters

PR #123, opened 2021-02-27
Type annotations in Display_text label library as a sample test for pylint and CI verifications

jposada202020:
As discussed during Saturday stream 2020-02-27 we want to include type annotations inside libraries code. Using this sample we want to verify CI behavior and possible solutions.

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -91,6 +91,7 @@ class Label(displayio.Group):
         if not max_glyphs and not text:
             raise RuntimeError("Please provide a max size, or initial text")
         if not max_glyphs:
+            text = "    ".join(text.split("\t"))
             max_glyphs = len(text)
         # add one to max_size for the background bitmap tileGrid
 
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -21,7 +21,7 @@ Implementation Notes
   https://github.com/adafruit/circuitpython/releases
 
 """
-
+from typing import Tuple
 import displayio
 
 __version__ = "0.0.0-auto.0"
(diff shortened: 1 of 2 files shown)


jposada202020:
As discussed: @hugodahl. Thanks let me know if you need help with anything :).
Nothing new here is falling since last week for the pylint bug. Thanks. Please review and comment.

adafruit_display_text/label.py
```diff
@@ -140,7 +140,7 @@ def __init__(
         if (anchored_position is not None) and (anchor_point is not None):
             self.anchored_position = anchored_position
 
-    def _create_background_box(self, lines, y_offset):
+    def _create_background_box(self, lines: int, y_offset: int):
```
hugodahl, line 143: Looks like one missed ` -> None` for the return type. All other locations already appear to have it added

adafruit_display_text/label.py
```diff
@@ -457,7 +456,7 @@ def anchored_position(self):
         )
 
     @anchored_position.setter
-    def anchored_position(self, new_position):
+    def anchored_position(self, new_position: Tuple[int, int]) -> Tuple[int, int]:
```
hugodahl, line 459: Since this is the setter for the `anchored_position` property, shouldn't the return type hint be `None`?

hugodahl (commented):
Just those two minor type-hint issues.

adafruit_display_text/label.py
jposada202020: Thanks for the revision, you are right!

adafruit_display_text/label.py
jposada202020: Yes, I miss that one, thanks for pointing that out!

jposada202020:
@hugodahl Thanks for the revision I made all the changes as suggested thanks for catching those up!!!

FoamyGuy:
This PR was superseded by #134 the type hints were included in that PR. The Tab replacement fix code was included with #132 

Thanks for working on this @jposada202020!

PR #128, opened 2021-03-07
Font examples

jposada202020:
Change in the example to include a font available in the font directory in the library


jposada202020:
will close #128

ladyada:
@FoamyGuy wanna take a look (you're the text/label guru! :)

FoamyGuy (commented):
Thanks @jposada202020! The new font and the example it's used in look good to me.

This does also contain the changes from #121. So when this is merged it will supersede #121.

We'll also need to wait on #126 also if we want the actions to pass.  I'll try to get that one merged tomorrow and then work on getting these others that aren't passing currently.

jposada202020:
@FoamyGuy No problem, Saturday I went to Free Fonts Rabbit hole, and the licenses in case you need something for the new guide. :). Yes. We would go back and majke the changes. I am planning to RE-submit all the PRs once the new library is approved

FoamyGuy:
These changes were included in #133

Issue #129, opened 2021-03-07
bitmap_label issue with lowercase letters


FoamyGuy:
I have discovered this example script (it's a trimmed down example from [bitmap_font](https://github.com/adafruit/Adafruit_CircuitPython_Bitmap_Font/blob/master/examples/bitmap_font_label_simpletest.py)):
```python
import board
from adafruit_display_text import bitmap_label as label
from adafruit_bitmap_font import bitmap_font

display = board.DISPLAY

font_file = "fonts/LeagueSpartan-Bold-16.bdf"


text = "HELLO WORLd"  #  ValueError
#text = "HELLO WORLD" #  works

font = bitmap_font.load_font(font_file)
color = 0xFF00FF

# Create the tet label
text_area = label.Label(font, text=text, color=color)

# Set the location
text_area.x = 20
text_area.y = 20

# Show it
display.show(text_area)

while True:
    pass
```

When the text has any lower case letters in it the bitmap_label will raise this exception:
```
Traceback (most recent call last):
  File "code.py", line 17, in <module>
  File "/lib/adafruit_display_text/bitmap_label.py", line 124, in __init__
  File "/lib/adafruit_display_text/bitmap_label.py", line 241, in _reset_text
  File "/lib/adafruit_display_text/bitmap_label.py", line 437, in _place_text
  File "/lib/adafruit_display_text/bitmap_label.py", line 461, in _blit
ValueError: out of range of target
```

The text with all upper case letters does render correctly however. label does not have the same issue, it will render both text versions fine. 

I'm not sure what the root cause of this issue is, perhaps something odd within this specific font? [The font is here](https://github.com/adafruit/Adafruit_CircuitPython_Bitmap_Font/blob/master/examples/fonts/LeagueSpartan-Bold-16.bdf)

Tested on:
```
Adafruit CircuitPython 6.2.0-beta.3-36-g103632f44 on 2021-03-07; Adafruit PyPortal with samd51j20
```
@kmatch98 I'm interested if you have any ideas about it this if you get a moment to check it out.

kmatch98:
The `blit` error message indicates that the code is trying to write pixels into a region outside of the target bitmap.  

Maybe the target bitmap size is wrong?  I'll take a look.

kmatch98:
[...]

PR #130, opened 2021-03-08
clip glyphs that exceed ascent property

kmatch98:
For font files that have `ASCENT`/`DESCENT` that is incompatible with the glyph dimensions and y-offsets, this clips the top of any offending glyphs and provides a warning.

This is to resolve https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/129

### DIFF
--- a/adafruit_display_text/bitmap_label.py
+++ b/adafruit_display_text/bitmap_label.py
@@ -488,7 +488,14 @@ class Label(displayio.Group):
                         y_blit_target = 0  # draw the clipped bitmap at y=0
 
                         print(
-                            'Warning: Glyph exceeds Ascent/Descent properties: "{}"'.format(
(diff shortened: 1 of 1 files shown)


kmatch98:
This failed the "similar lines" pylint error.  @foamyguy or @kattni, can you install the proper incantations to clear that hurdle.

FoamyGuy (approved):
This make sense to me. I tested out this solution and it does allow the font used in that issue to render properly now without crashing.

Looks good to me.

As for the duplicate pylint check I think that will get resolved when #126 is merged. I will try to finish that up today and see if we can get it merged. If we get that one merged I think a few of these other PRs will start passing actions. Though some of them may require some tweaks to fit in the refactored classes from #126.

kmatch98:
Done.  Should be good to go.  Thanks for the refactor that made this one go smoothly.

PR #132, opened 2021-03-10
Tab replacement corrections

jposada202020:
To replace PR #121. Testing code is the same as used in previous PR

Tested with bitmap_label and label. Similar results as PR #121 showed.  Included corrections from @kmatch98 in that PR.

image

@FoamyGuy as discussed please review.


jposada202020:
So happy to finally see a Green check!!

FoamyGuy (approved):
These changes look good to me. Thank you @jposada202020 for all our work on this fix!

I retested the example code from the original PR successfully with this newest version on:
```
Adafruit CircuitPython 6.2.0-beta.3-36-g103632f44 on 2021-03-07; Adafruit PyPortal with samd51j20
```

PR #133, opened 2021-03-10
New font

FoamyGuy:
This contains the one commit from #128 that adds the new font and updated example. One more commit to tweak the license filename to match.

updated example and font was tested on: 
```
Adafruit CircuitPython 6.2.0-beta.3-36-g103632f44 on 2021-03-07; Adafruit PyPortal with samd51j20
```


jposada202020:
I tested this using the Example code

image

FoamyGuy:
Thanks for testing this out @jposada202020.

PR #138, opened 2021-03-12
Create a new Feature to label to add syles capabilities

jposada202020:
## Features
The change will add the feature to use label with styles. With this feature you could possible create the same style among differents widgets. The syles are taken from the PySimpleGui Project

## Changes
There are changes in the following files:
__init__: A new function is added. This function converts the color `str` representation used in colorset used in the source library
label.py: a New class LabelT is added. This class is a child of label, This could use the theme as a parameter.
styles.py: New file including the colorsets. the idea to create a new file will be to share among different widgets. This could be easily done adding entries to the dictionary.
display_text_label_styles.py: new example

## Tests
Expected result of the example is:
image

## Others
Thanks to @qwertimer and @LewsTherinTelamon for the ideas, feel free to comment and discuss
@PaintYourDragon if you could guide us in the Colorset used by Adafruit that would be really helpful.

### DIFF
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -144,6 +144,12 @@ def wrap_text_to_lines(string, max_chars):
     return the_lines
 
 
+def get_hex(color: str) -> int:
+    """Function to convert Color styles to HEX
+    :param str color: string representation of Hex color"""
+    return int(color[1:], 16)
+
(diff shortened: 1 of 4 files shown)


jposada202020:
This is closed. Create a new Pull Request, pushing on my side create a new PR.

Issue #140, opened 2021-03-13
anchored position issue


FoamyGuy:
Issue and test code outlined here: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/pull/137#issuecomment-795956051

jposada202020:
@FoamyGuy I have retested this, with the new refactor. This is working as expected. There is A mistake in my test code.  I test this in the pyportal and in the Pygame Display without any problems
The problem was in the setter as I was not using the `=` sign. If you coudl retest  it and verify that everything is working
```python

import terminalio
import displayio
from os import uname
import time

if uname()[0] == 'samd51':
    import board
else:
    from blinka_displayio_pygamedisplay import PyGameDisplay
from adafruit_display_text import label, bitmap_label
from adafruit_bitmap_font import bitmap_font

if uname()[0] == 'samd51':
    display = board.DISPLAY
else:
    display = PyGameDisplay(width=320, height=240)
splash = displayio.Group(max_size=10)
MEDIUM_FONT = bitmap_font.load_font("fonts/Helvetica-Bold-16.bdf")
LINE_HOR = displayio.Bitmap(4, 320, 2)
LINE_VER = displayio.Bitmap(320, 4, 2)
palette = displayio.Palette(2)
palette[0] = 0x004400
palette[1] = 0x00FFFF


text = "CircuitPython"
text_area = bitmap_label.Label(MEDIUM_FONT,
                        text="First Text",
                        label_direction="LTR",
                        background_tight=True,
                        x=200,
                        y=100,
                        padding_left=10,
                        padding_top=10,
                        padding_bottom=10,
                        padding_right=10,

                               )
text_area.anchored_position=(200, 100)
text_area.anchor_point=(0.5, 0.5)
splash.append(text_area)


text_area_UPR = bitmap_label.Label(MEDIUM_FONT, text=text,
                        # label_direction="UPR",
                        x=10,
                        y=10,)

splash.append(text_area_UPR)


text_area_DWR = bitmap_label.Label(MEDIUM_FONT,
                        text="Upwards",
                        label_direction="UPR",
                        # x=200,
                        # y=200,
                        anchored_position=(200, 200),
                        anchor_point=(0.5, 1.0),

)

line_ver= displayio.TileGrid(LINE_VER,
                               pixel_shader=palette,
                               x=0,
                               y=200)
splash.append(line_ver)
line_hor= displayio.TileGrid(LINE_HOR,
[...]

PR #145, opened 2021-03-19
initial Commit for draft Bitmap Label directional label

jposada202020:
@kmatch98 I will need your help here, as under the hood labels are different, If we want to implement directional label for the bitmaplabel, we need to do a different approach as we do not have the same tools, what would you recommend? do we need a rotation matrix, we need do `-1` and also a rotation.

Thanks 🧑‍🔬


kmatch98:
Probably the simplest approach is to generate the `bitmap_label` bitmap as usual and then do some transformation into the final bitmap.

Here are two approaches I can think of:
1.  Use the `bitmaptools.rotozoom` function to do a rotational "blit" into a new bitmap.  (This function can do any angle you want.) You will need to calculate the size of the final "rotated" bitmap before you perform the "rotozoom" into it.  If you are only allowing 90 degree increments, the the bitmap size calculation is easy. One downside is there could be some integer rounding errors that could make the final result not match the original bitmap.

2. Write a bitmap transformation function to do the 90 degree angles.  Since you are only doing 90 degree angles it is much simpler math to calculate each pixel's position.  And you can write this (to avoid rounding erros) to ensure the rotated bitmap exactly matches the original.

If you want to get more complex, you could modify the character-by-character blitting function and then adjust the rotation and position of each character individually.  This more complex approach would probably use the lowest amount of RAM during the process, since you won't need two bitmaps (like #1 and #2 above).

Let me know your thoughts or if you'd like other suggestions.

kmatch98:
Oh!  I just remembered that `TileGrid` has a rotation function.  Since the label is in a `TileGrid`, it's probably easiest to just manipulate that.

jposada202020:
@kmatch98 Thanks for the info.
For the RTL (Right-to-Left) text and TTB Top-To-Bottom I think I will use the same approach that I used in label. You could used in release 2.17.0
I think that for text upwards and downwards text I will explore the `rotozoom` option as you mentioned. I just want to preserve the memory efficiency and speed of the bitmap_label and I did not wanted to do a "Tilegrid" conversion, as this is in every glyph and the the bitmap one. Thanks for pointing me in the right direction you saved me a lot of time :). 
But I think it will be a win if at least people will have the Upwards and Downwards feature not too slow.
Thanks

jposada202020:
Four directions are implemented here similar to label:
* RTL
* LTR
* DWR
* UPR

I am not seeing an easy way to implement TTB, so I think is better to maintain the memory efficiency and if people need text from top to bottom they can use label.  Still having problems when anchoring positions, I try to troubleshoot a little but with no luck.  I tried version 2.14 of the library and bitmap_label anchor characteristics were working. #140 . I am guessing that we are having problems in the setter, because we are receiving a None as a anchor_position.

image

@kmatch98 thanks for the suggestions, as you can see it make this a lot easier.

kmatch98:
I evaluated this and it works as expected.  

Right now you are using a different method to achieve left-to-right "LTR" versus the other options. 
I recommend doing all of the transformation using just the tilegrid options (`flip_x`, `flip_y` and `transpose_xy`).

You can achieve "LTR" with `self.tilegrid.flip_x=True`.
Also, if you want to achieve upside down, you can use `self.tilegrid.flip_x = self.tilegrid.flip_y = True`.

Otherwise, it looks good to me.

jposada202020:
@kmatch98 Thank you for the review, as previously mentioned thanks again for point me in the right direction, it was a breeze compared to do the same feature in label after your indications.
Regarding the LTR,  just to verify that I am understanding correctly, when you said flip_x I think that is more related to "RTL" that I used a string method. If so there is a difference in this

`PA` flip_x = `Aꟼ`
`PA` "".join(reversed(PA)) = `AP`

For RTL languages we write normal just in the Left-to-right direction, and not mirroring the letters, that is why methods are different.
Could you confirm your confirm that I am understanding correctly.

For the downwards, thank it was something that I discover when doing the PR, I was considering adding, but label does not have it, that is why I did not include it.

kmatch98:
Sorry that was a typo. You are right that it is RTL. Keep up the good work. When you are ready I will take another look.

jposada202020:
@kmatch98 what a ride :). As of you just a single little tiny, small tiny little space..... lol and a rabbit whole , anyway I include the UPD -Upside Down. Thanks.

FoamyGuy:
I think we need to move this check for valid direction out of LabelBase and put it into each of the subclasses since they have slightly different options with `UPD` and `TTL`.

https://github.com/jposada202020/Adafruit_CircuitPython_Display_Text/blob/69909e75f3451fa1aa2730c635f95aaafaa5db79/adafruit_display_text/__init__.py#L239

I tried making an upside down label like this
```
text_area = bitmap_label.Label(terminalio.FONT, text=text, label_direction="UPD")
```
but I got this error: 
```
code.py output:
  File "code.py", line 10, in <module>
  File "/lib/adafruit_display_text/bitmap_label.py", line 91, in __init__
  File "/lib/adafruit_display_text/__init__.py", line 239, in __init__
RuntimeError: Please provide a valid text direction
```

jposada202020:
@FoamyGuy good point thank you for reviewing, I'll fix that, Thanks

docs/examples.rst
```diff
@@ -6,3 +6,30 @@ Ensure your device works with this simple test.
...
+Example to compare Label and Bitmap_Label characteristics
+
+.. literalinclude:: ../examples/display_text_label_vs_bitmap_label_comparison.py
+    :caption: examples/display_text_label_vs_bitmap_label_comparison.py
+    :linenos:
+
+Label vs Bitmap_label Comparison
+--------------------------------
+
+Example to compare Label and Bitmap_Label characteristics
```
FoamyGuy, line 31: The code changes are looking good to me now. One more small cleanup item in the examples docs and then I think this will be ready. 

It looks like this `Label vs Bitmap_Label Comparison` example got doubled up in this `examples.rst` file. There are two copies of it linking to the same example. Lets remove one of them.

jposada202020: @FoamyGuy Thanks, you are rigt, I corrected here, however all the exemplaes are in in the API examples modifications for #147.

FoamyGuy (approved):
Looks good to me thanks for all of your work on this @jposada202020

PR #149, opened 2021-04-01
Ascent_Descent information at start

jposada202020:
This will make info on the font ascent/descent info available for the class, avoiding unnecessary calls to calculate this during the script

### DIFF
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -245,7 +245,7 @@ class LabelBase(Group):
         if self.base_alignment:
             self._y_offset = 0
         else:
-            self._y_offset = self._get_ascent() // 2
+            self._y_offset = self._ascent // 2
 
     def _get_ascent_descent(self) -> Tuple[int, int]:
         """ Private function to calculate ascent and descent font values """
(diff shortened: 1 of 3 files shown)


lesamouraipourpre (changes requested):
Looking good.

You could also remove  the `_get_ascent` function and it's only call:
```
diff --git a/adafruit_display_text/__init__.py b/adafruit_display_text/__init__.py
index e6f2c1d..246f54e 100644
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -245,7 +245,7 @@ class LabelBase(Group):
         if self.base_alignment:
             self._y_offset = 0
         else:
-            self._y_offset = self._get_ascent() // 2
+            self._y_offset = self._ascent // 2
 
     def _get_ascent_descent(self) -> Tuple[int, int]:
         """ Private function to calculate ascent and descent font values """
@@ -268,9 +268,6 @@ class LabelBase(Group):
                 descender_max = max(descender_max, -this_glyph.dy)
         return ascender_max, descender_max
 
-    def _get_ascent(self) -> int:
-        return self._get_ascent_descent()[0]
-
     @property
     def font(self) -> None:
         """Font to use for text display."""
```

jposada202020:
@lesamouraipourpre Good point will do. Thanks for reviewing :) and the recommendation

lesamouraipourpre (approved):
That's a nice simplification you've implemented.

FoamyGuy (approved):
Looks good to me, thanks @jposada202020. 

And thank you @lesamouraipourpre for testing!

PR #150, opened 2021-04-02
Library Advanced example and tests

jposada202020:
Adding a new advanced example that includes all the parameters features.

### DIFF
--- a/examples/display_text_advance_example.py
+++ b/examples/display_text_advance_example.py
@@ -501,7 +501,7 @@ except TypeError:
         background_color=0x004499,
     )
     main_group.append(warning_text)
-    time.sleep(TIME_PAUSE))
+    time.sleep(TIME_PAUSE)
     display.show(main_group)
 
 main_group.pop()
--- a/examples/display_text_advance_example.py
+++ b/examples/display_text_advance_example.py
@@ -1,0 +1,640 @@
+# SPDX-FileCopyrightText: 2021 Jose David M.
+#
+# SPDX-License-Identifier: MIT
+#############################
+"""
+This is an advanced demonstration of the display_text library capabilities
+"""
+
+import time
+import board
+import displayio
+import terminalio
+import fontio
+from adafruit_bitmap_font import bitmap_font
+from adafruit_display_text import label, bitmap_label
+
+display = board.DISPLAY
+main_group = displayio.Group(max_size=10)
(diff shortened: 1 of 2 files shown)


examples/display_text_advance_example.py
```diff
@@ -0,0 +1,640 @@
...
+import board
+import displayio
+import terminalio
+import fontio
+from adafruit_bitmap_font import bitmap_font
+from adafruit_display_text import label, bitmap_label
+
+display = board.DISPLAY
+main_group = displayio.Group(max_size=10)
+font_name = "fonts/LibreBodoniv2002-Bold-10.bdf"
```
lesamouraipourpre, line 19: font_name isn't used. It can safely be removed.

examples/display_text_advance_example.py
```diff
@@ -0,0 +1,640 @@
...
+MEDIUM_FONT = bitmap_font.load_font("fonts/LibreBodoniv2002-Bold-10.bdf")
```
lesamouraipourpre, line 20: The font LibreBodoniv2002-Bold-10.bdf isn't included in the package, only the size 27 one is.

lesamouraipourpre (commented):
Other than the missing font (I substituted something else for testing) this is looking really good cycling through all the possibilities.

Tested on a PyPortal Pynt.

jposada202020:
@lesamouraipourpre Thanks, I will work on that, thanks for reviewing and the comments. I will change the font a rework on the name. Thanks

examples/display_text_advance_example.py
jposada202020: I remove the font_name reference as suggested

examples/display_text_advance_example.py
jposada202020: Thanks I change the font to include the one present in the repo
```python
MEDIUM_FONT = bitmap_font.load_font("fonts/LeagueSpartan-Bold-16.bdf")
```
This change was done in https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/pull/150/commits/0e385c186c9c57e8528bbeba58a8be96e6be732d

lesamouraipourpre (approved):
Retested on a PyPortal Pynt and it works great.

FoamyGuy (approved):
Looks good to me 

tested successfully on:
```
Adafruit CircuitPython 6.2.0-rc.0 on 2021-04-01; Adafruit PyPortal with samd51j20
```

PR #153, opened 2021-07-01
Remove max_size parameter

### LINKED ISSUE #131: Try to remove max_glyphs restriction from label
Making an issue now so that we keep this in mind. I do think we'll want to hold off on making this change at least until there is a stable release made that includes the changes to Group though so it might be a little bit until we actually need to do anything further.

displayio.Group has been updated to allow it's size to be mutable so more things can be added, and there is no need to use the `max_size` any longer.

We will be able to change label.Label to make use of this and remove the `max_glyphs` restriction.

lesamouraipourpre:
Closes #131

Remove the `max_size` parameter from `Label` and `BitmapLabel` now that it has been disabled in `shared-bindings`.

This has been tested with:
```
Adafruit CircuitPython 6.3.0 on 2021-06-01; Adafruit PyPortal with samd51j20
```

### DIFF
--- a/adafruit_display_text/label.py
+++ b/adafruit_display_text/label.py
@@ -81,18 +81,15 @@ class Label(LabelBase):
         text = kwargs.get("text", "")
 
         if not max_glyphs and not text:
-            raise RuntimeError("Please provide a max size, or initial text")
+            raise RuntimeError("Please provide a max_glyphs, or initial text")
         self._tab_replacement = kwargs.get("tab_replacement", (4, " "))
         self._tab_text = self._tab_replacement[1] * self._tab_replacement[0]
         text = self._tab_text.join(text.split("\t"))
         if not max_glyphs:
             max_glyphs = len(text)
-        # add one to max_size for the background bitmap tileGrid
 
         # local_group will set the scale
-        self.local_group = displayio.Group(
-            max_size=max_glyphs + 1, scale=kwargs.get("scale", 1)
-        )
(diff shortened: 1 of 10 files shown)


adafruit_display_text/label.py
```diff
@@ -81,18 +81,15 @@ def __init__(self, font, **kwargs) -> None:
         text = kwargs.get("text", "")
 
         if not max_glyphs and not text:
-            raise RuntimeError("Please provide a max size, or initial text")
+            raise RuntimeError("Please provide a max_glyphs, or initial text")
```
FoamyGuy, line 84: I think we may be able to get rid of `max_glyphs`. if I understand correctly, the reason for that restriction and parameter was the underlying Group's `max_size`.

lesamouraipourpre: It's assigned to width further down so I didn't want to mess with it for now.
https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/1f27ba7554b789f5d987daff652df0430dbcb483/adafruit_display_text/label.py#L98

FoamyGuy: Interesting. I'll test out this change as-is, we can merge it now if good and then remove it in a followup.

it seems odd that it would use that for width without multiplying by the width of the glyph in pixels. I can poke into it a bit and try to figure out if it's needed or how to replace it with something different.

FoamyGuy (approved):
These changes look good to me. I tested successfully a handful of the examples on a PyPortal with CircuitPython 6.3.0 and 7.0.0.beta-3.

PR #154, opened 2021-07-10
Refactor the classes

lesamouraipourpre:
The aim of this refactor is as follows:

- Deduplicate any code in the subclasses and move it up to `LabelBase`
- Where code only applies to one of the subclasses, have no knowledge of it in the other subclass or the base class.
- Remove `max_glyphs` - the removal of `max_size` from `displayio.Group` makes it redundant. - https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/131
- Make all variables private/protected (start with an _). **Note:** This may break any external code which relied on access to these.

These changes have been tested on a PyPortal. I can not see any problems with the examples but it's possible I'm missing a fault.


jposada202020:
Just to open the discussion here, for me I would keep some of the variables public:
* anchor_position
* anchored_position
* base_alignment
* label_direction
As any change in these variables will change library behavior. However, this is just a personal(functional) preference. So no problem either way.
Also, you mentioned that you test the examples, so I guess we would cover a lot of the example code in the learning guides.

lesamouraipourpre:
`anchor_point`, `anchored_position` & `label_direction` are all available as properties of `LabelBase`
`base_alignment` is currently only a constructor parameter but could easily be made into a property as well.

jposada202020:
good it is solved then  could do PR for the basealignment after :)

FoamyGuy:
The PyPortal example: https://github.com/lesamouraipourpre/Adafruit_CircuitPython_Display_Text/blob/refactor/examples/display_text_pyportal.py

crashes with this traceback on the branch from this PR:
```
  File "code.py", line 41, in <module>
AttributeError: 'Label' object has no attribute 'height'
```
I'm working on testing the rest of the examples as well.

FoamyGuy:
I tested all of the other examples and all are working as intended. I looked over these changes and everything looks good to me beyond the `height` property issue mentioned above.

The only other thought that I had is it might be worthwhile to temporarily (for a few months / release versions) check if user code passed in a `max_glyphs` and if so print a warning message that tells them it's no longer necessary and can be safely removed. I'm not 100% sure if we do want to do this though as it could lead to extra "spam" being printed in the serial console and there does not seem to be any harm in user code passing it in with this version, it just gets ignored. It may be worth asking some more of the team members to see if folks think it's a good idea or not. 

Thanks for working on this @lesamouraipourpre!

lesamouraipourpre:
I've added `height` and `width` properties to `LabelBase`. The previously defined `height` and `width` were defined differently on `Label` and `BitmapLabel`. They are now defined on the root class based on the `bounding_box`.

`max_glyphs` - I've added a printed message "Please update your code: 'max_glyphs' is not needed anymore."

FoamyGuy (approved):
Looks good to me. I Re-tested the `display_text_pyportal.py` example successfully after the latest commits and all is well.

Thanks again @lesamouraipourpre

Issue #157, opened 2021-08-19
save_text=False throws error on init


sargunv:
```py
label = Label(
    font=FONT,
    color=color,
    anchor_point=(0.0, 0.0),
    anchored_position=(column, 0),
    text=str,
    save_text=False,
)
```

```
  File "adafruit_display_text/bitmap_label.py", line 100, in __init__
  File "adafruit_display_text/bitmap_label.py", line 161, in _reset_text
  File "adafruit_display_text/bitmap_label.py", line 277, in _text_bounding_box
TypeError: 'NoneType' object is not iterable
```

I believe this is because that `_reset_text` function is using the property `self._text` instead of the local variable `text` in some (but not all) places.

lesamouraipourpre:
I am unable to reproduce this on a PyPortal running `Adafruit CircuitPython 6.3.0 on 2021-06-01` or `Adafruit CircuitPython 7.0.0-alpha.6 on 2021-08-12`

**What version of CircuitPython are you running?**
This will be in the contents of `boot_out.txt` on the device.

**What versions of the libraries are installed?**
Run `circup freeze` from the connected computer. If you need to install `circup`, run `pip3 install circup`
For example, I get
```bash
> circup freeze
Found device at /media/user/CIRCUITPY, running CircuitPython 7.0.0-alpha.6.
adafruit_fakerequests==1.0.3
adafruit_miniqr==1.3.6
adafruit_pixelbuf==1.1.0
adafruit_requests==1.10.0
adafruit_touchscreen==1.1.6
neopixel==6.2.1
simpleio==3.0.1
adafruit_bitmap_font==1.5.1
adafruit_bus_device==5.0.6
adafruit_display_text==2.20.0
adafruit_esp32spi==3.5.11
adafruit_io==5.5.0
adafruit_portalbase==1.9.1
adafruit_pyportal==6.0.0
```

You can use `circup update` to update the libraries on the device to the latest version if needed.

tekktrik:
Hi @sargunv, I noticed your code has the parameter `text=str`.  Is it possible you're setting the text to the type `str` then an actual string value.  That is to say does it work if you use something like `text="LabelName"`?

tekktrik:
Or were you just saying that any string causes the issue?

tekktrik:
I ask because I'd love to look into this if you have any more information :)

PaulskPt:
[...]

PR #160, opened 2021-12-21
Fix error from save_text = False

tekktrik:
This PR should address Issue #157 but I don't have the hardware to test, just followed the code to figure out what was happening.  Use of local variable `text` instead of `self._text` should fix the issue.

### DIFF
--- a/adafruit_display_text/bitmap_label.py
+++ b/adafruit_display_text/bitmap_label.py
@@ -157,7 +157,7 @@ class Label(LabelBase):
                 loose_box_y,
                 loose_y_offset,
             ) = self._text_bounding_box(
-                self._text,
+                text,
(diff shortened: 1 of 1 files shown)


PaulskPt:
I applied the changes to bitmap_label.py. As far as I can see it works. No crash. See the annotated screenshot.

adafruit_display_text__bitmap_label py_mod_by_tekktrik

tekktrik:
Yahoo!  That's great to hear!

FoamyGuy (approved):
Looks good to me. Thanks @tekktrik for the fix and @PaulskPt for trying it out.

tekktrik:
My pleasure, thanks everyone for the help and support in addressing the issue!

PR #161, opened 2021-12-21
Update type hints

tekktrik:
Found some type hints were missing and some docstrings could use some tweaks, so just a PR to address those!


FoamyGuy (approved):
LGTM thanks @tekktrik

PR #164, opened 2022-02-08
Adding ScrollingLabel

FoamyGuy:
This adds a new `ScrollingLabel` that is a fixed-width label that can scroll the text leftward like a marquee.

An example has been provided to show the most basic usage.

This component is used in my PyPortal Winamp player project but I figured this piece is generic enough to re-used for other purposes.

### DIFF
--- a/adafruit_display_text/scrolling_label.py
+++ b/adafruit_display_text/scrolling_label.py
@@ -38,7 +38,7 @@ class ScrollingLabel(bitmap_label.Label):
 
     :param font: The font to use for the label.
     :param max_characters: The number of characters that sets the  fixed-width. Default is 10.
-    :param full_text: The full text to show in the label. If this is longer than
+    :param text: The full text to show in the label. If this is longer than
      `max_characters` then the label will scroll to show everything.
     :param animate_time: The number of seconds in between scrolling animation
      frames. Default is 0.3 seconds.
--- a/adafruit_display_text/scrolling_label.py
+++ b/adafruit_display_text/scrolling_label.py
@@ -1,0 +1,149 @@
+# SPDX-FileCopyrightText: 2019 Scott Shawcroft for Adafruit Industries
+#
+# SPDX-License-Identifier: MIT
+
+"""
+`adafruit_display_text.scrolling_label`
+====================================================
+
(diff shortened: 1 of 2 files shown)


kmatch98:
- I verified that `label_direction` works as expected giving different rotation options.
- I verified that `background_color` works.  On proportional fonts, the right side of the label gets resized as it scrolls, that's just a consequence of the backside bounding box changing sizes.  That could be an area for future improvement.
- Redefining `full_text` works too.

#### One item for consideration:
The keyword argument `full_text` had to be used so that you wouldn't conflict with the existing superclass `.text` property.  And I see you put a warning, but it seems a little harder to use that way.

Perhaps making the this a subclass `Group` (that contains your Label) would give you the flexibility to use an input parameter named `text`.  This will make its external definition more similar to Label and then easier to convert between Label and ScrollingLabel.

FoamyGuy:
Thanks for taking a look. I do like this idea better to use `text` instead of `full_text` and switch the parent class to allow for it. I'll make that change tonight and add a new commit with it.

FoamyGuy:
After tinkering with it for a bit I'm interested to see if there is a way to do this without having to extend Group instead of Label. If we change the super class it means we would lose all of the "pass through" properties so for setting things like `color` or `background_color` get tricky.

Something like this would no longer work:

`my_scrolling_label.color = 0x00ff00`

we could add "boilerplate" properties that copy all of the Label ones and pass them through to the Label instance but I wonder if there is some way we can avoid needing all of the extra code to re-declare each property.


I have changed the argument name to `text` instead of `full_text` and the warning. But it does kind of leave open the issue that user might get unexpected results if they are changing the text after the fact trying to do something like:

`my_scrolling_label.text = "the new text long enough to scroll"`

kmatch98:
Thanks for the comments about using Label versus Group, and it makes sense.  It's awkward to initialize with `text` input parameter and then `full_text` as a getter/setter but I understand the limit.  I remember previously trying to overriding Group's superclass instance variables versus subclass variables and hit one of the limits of CircuitPython (I think it was with scaling).

I think this is ok for now, so I'll merge it.

PR #165, opened 2022-02-18
Add missing adafruit-circuitpython-bitmap-font dependency

tekktrik:
Now used in type annotations


FoamyGuy (approved):
Looks good to me. Thanks for the fix @tekktrik

PR #167, opened 2022-03-12
Improved memory efficiency for some text updates to bitmap_label.Label

kevinjwalters:
1. `bitmap_label.Label` - the `self._bitmap` and `self._tilegrid` is no longer recreated all the time
  1. This makes it more efficient for updates for text have the _same length_ particularly in terms of avoiding creating a new `Bitmap` (and `TileGrid`) using more memory briefly and potentially causing memory fragmentation.
  2. Adding a new `self._prev_label_direction` to support `label_direction` properly.
4. Fixed an unrelated bug in `label_direction` for `RTL` which would toggle direction if the value was repeatedly set - text reversal is now performed as text is set and not stored.
5. `_replace_tabs()` function also enhanced to only do split/join if tabs are present - this is probably a rare case and it's inefficient to be making new text strings all the time. 

The most efficient updates are now if the text string is the same character length as the previous value. The bitmap is cleared with `fill(0)` and re-written and this can cause flicker. The cognoscenti will be controlling display refresh, of course.

### DIFF
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -283,7 +283,7 @@ class LabelBase(Group):
             self._y_offset = self._ascent // 2
 
     def _get_ascent_descent(self) -> Tuple[int, int]:
-        """ Private function to calculate ascent and descent font values """
+        """Private function to calculate ascent and descent font values"""
         if hasattr(self.font, "ascent") and hasattr(self.font, "descent"):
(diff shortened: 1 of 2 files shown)


FoamyGuy (approved):
Looks good to me. Thanks for the improvements @kevinjwalters!

Tested successfully with PyPortal and some variations of the examples scripts.

PR #168, opened 2022-04-02
Free the bitmap when set to empty string

Neradoc:
In the latest release, setting the text to an empty string, then back to a string of the same size as the previous (non-empty) value, the bitmap label would not be updated, remaining invisible until its size changed. [See this code for example](https://github.com/adafruit/Adafruit_CircuitPython_MacroPad/blob/main/examples/macropad_grid_layout.py). This is because `self._bitmap` was still set, with the right size, so was `self._tilegrid`, so `self._local_group` remained empty.

(Also one use of `self._text` instead of `text` fixed).

### DIFF
--- a/adafruit_display_text/bitmap_label.py
+++ b/adafruit_display_text/bitmap_label.py
@@ -154,6 +154,10 @@ class Label(LabelBase):
             for _ in self._local_group:
                 self._local_group.pop(0)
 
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Thanks! This makes sense to me. Assuming you tested. Making a release immediately.

PR #170, opened 2022-05-11
Add typing and documentation for ScrollingLabel

rgrizzell:
I noticed it was missing.

### DIFF
--- a/adafruit_display_text/scrolling_label.py
+++ b/adafruit_display_text/scrolling_label.py
@@ -73,7 +73,7 @@ class ScrollingLabel(bitmap_label.Label):
 
         self.update()
 
-    def update(self, force: Optional[bool] = False) -> None:
+    def update(self, force: bool = False) -> None:
         """Attempt to update the display. If ``animate_time`` has elapsed since
         previews animation frame then move the characters over by 1 index.
         Must be called in the main loop of user code.
@@ -122,7 +122,7 @@ class ScrollingLabel(bitmap_label.Label):
         return self._current_index
 
     @current_index.setter
-    def current_index(self, new_index: int):
+    def current_index(self, new_index: int) -> None:
         if new_index < len(self.full_text):
             self._current_index = new_index
         else:
@@ -138,7 +138,7 @@ class ScrollingLabel(bitmap_label.Label):
         return self._full_text
 
     @full_text.setter
-    def full_text(self, new_text: str):
+    def full_text(self, new_text: str) -> None:
         if new_text[-1] != " ":
             new_text = "{} ".format(new_text)
         self._full_text = new_text
--- a/adafruit_display_text/scrolling_label.py
+++ b/adafruit_display_text/scrolling_label.py
@@ -27,10 +27,8 @@ __version__ = "0.0.0-auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
 
 try:
-    from typing import Union, Optional
-    from fontio import BuiltinFont
-    from adafruit_bitmap_font.bdf import BDF
-    from adafruit_bitmap_font.pcf import PCF
+    from typing import Optional
+    from fontio import FontProtocol
 except ImportError:
     pass
 
@@ -43,7 +41,7 @@ class ScrollingLabel(bitmap_label.Label):
     in order to show the full text if it's larger than the fixed-width.
 
     :param font: The font to use for the label.
-    :type font: ~BuiltinFont, ~BDF, or ~PCF
+    :type: ~FontProtocol
     :param int max_characters: The number of characters that sets the fixed-width. Default is 10.
     :param str text: The full text to show in the label. If this is longer than
      ``max_characters`` then the label will scroll to show everything.
@@ -55,8 +53,8 @@ class ScrollingLabel(bitmap_label.Label):
     # pylint: disable=too-many-arguments
     def __init__(
         self,
-        font: Union[BuiltinFont, BDF, PCF],
-        max_characters: Optional[int] = 10,
+        font: FontProtocol,
+        max_characters: int = 10,
         text: Optional[str] = "",
         animate_time: Optional[float] = 0.3,
         current_index: Optional[int] = 0,
@@ -135,7 +133,7 @@ class ScrollingLabel(bitmap_label.Label):
         """The full text to be shown. If it's longer than ``max_characters`` then
         scrolling will occur as needed.
 
-        :return string: The full text of this label.
+        :return str: The full text of this label.
         """
         return self._full_text
 
--- a/adafruit_display_text/scrolling_label.py
+++ b/adafruit_display_text/scrolling_label.py
@@ -26,36 +26,42 @@ Implementation Notes
 __version__ = "0.0.0-auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
 
+try:
+    from typing import Union, Optional, Tuple
+    from fontio import BuiltinFont
+    from adafruit_bitmap_font.bdf import BDF
+    from adafruit_bitmap_font.pcf import PCF
+except ImportError:
+    pass
+
 import time
 from adafruit_display_text import bitmap_label
 
 
 class ScrollingLabel(bitmap_label.Label):
-
-    """
-    ScrollingLabel - A fixed-width label that will scroll to the left
+    """ScrollingLabel - A fixed-width label that will scroll to the left
     in order to show the full text if it's larger than the fixed-width.
 
     :param font: The font to use for the label.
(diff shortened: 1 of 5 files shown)


adafruit_display_text/scrolling_label.py
```diff
@@ -26,36 +26,42 @@
 __version__ = "0.0.0-auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
 
+try:
+    from typing import Union, Optional
+    from fontio import BuiltinFont
```
tekktrik, line 31: Don't have the time to give a full check this instant, but something I noticed that I wanted to mention so this doesn't get merged :)

`fontio` should have `FontProtocol` in it that was created for type annotations to take the place of `Union[BuiltinFont, BDF, PCF]` so we should use that!  I'll double check that this is indeed valid for these type annotations.

adafruit_display_text/scrolling_label.py
```diff
@@ -26,36 +26,42 @@
...
+    from adafruit_bitmap_font.bdf import BDF
+    from adafruit_bitmap_font.pcf import PCF
```
tekktrik, line 33: You can change this over to use the `fontio.FontProtocol` typing protocol!

```suggestion
    from fontio import FontProtocol
```

adafruit_display_text/scrolling_label.py
```diff
@@ -26,36 +26,42 @@
 __version__ = "0.0.0-auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
 
+try:
+    from typing import Union, Optional
```
tekktrik, line 30: You won't need `Union` anymore if you use `fontio.FontProtocol` either!

```suggestion
    from typing import Optional
```

adafruit_display_text/scrolling_label.py
```diff
@@ -26,36 +26,42 @@
...
 
     # pylint: disable=too-many-arguments
     def __init__(
         self,
-        font,
-        max_characters=10,
-        text="",
-        animate_time=0.3,
-        current_index=0,
+        font: Union[BuiltinFont, BDF, PCF],
```
tekktrik, line 58: This can change over to `fontio.FontProtocol`:

```suggestion
        font: FontProtocol,
```

adafruit_display_text/scrolling_label.py
```diff
@@ -26,36 +26,42 @@
...
+    from adafruit_bitmap_font.pcf import PCF
+except ImportError:
+    pass
+
 import time
 from adafruit_display_text import bitmap_label
 
 
 class ScrollingLabel(bitmap_label.Label):
-
```
tekktrik, line 34: Thank you :pray:

adafruit_display_text/scrolling_label.py
```diff
@@ -26,36 +26,42 @@
...
-    ScrollingLabel - A fixed-width label that will scroll to the left
+    """ScrollingLabel - A fixed-width label that will scroll to the left
     in order to show the full text if it's larger than the fixed-width.
 
     :param font: The font to use for the label.
-    :param max_characters: The number of characters that sets the  fixed-width. Default is 10.
-    :param text: The full text to show in the label. If this is longer than
-     `max_characters` then the label will scroll to show everything.
-    :param animate_time: The number of seconds in between scrolling animation
+    :type font: ~BuiltinFont, ~BDF, or ~PCF
```
tekktrik, line 46: This can also change to `fontio.FontProtocol`:

```suggestion
    :type: ~FontProtocol
```

adafruit_display_text/scrolling_label.py
```diff
@@ -26,36 +26,42 @@
...
     # pylint: disable=too-many-arguments
     def __init__(
         self,
-        font,
-        max_characters=10,
-        text="",
-        animate_time=0.3,
-        current_index=0,
+        font: Union[BuiltinFont, BDF, PCF],
+        max_characters: Optional[int] = 10,
```
tekktrik, line 59: So `Optional` is actually a horrible name for what it describes :sweat_smile:  What it denotes is something that can take `None` as a value.  For example if I had an argument for a class that could set the color for something as a hex code, or `None` to signify that it should be transparent, it would be:

```python
def set_color(color: Optional[int]):
    ...
```

It looks like these can't take `None` as a value, so despite being optional to provide, they aren't `Optional`.

In this case, despite being optional to provide, the argument still is/expects to be a `str` type, so it would just be:

```suggestion
        text: str = "",
```

Same for the ones below!

adafruit_display_text/scrolling_label.py
```diff
@@ -69,13 +75,13 @@ def __init__(
 
         self.update()
 
-    def update(self, force=False):
-        """
-        Attempt to update the display. If `animate_time` has elapsed since
+    def update(self, force: Optional[bool] = False) -> None:
```
tekktrik, line 78: See the previous comment about `Optional`.  `force` should be type annotated as just `bool`.

adafruit_display_text/scrolling_label.py
```diff
@@ -110,33 +116,31 @@ def update(self, force=False):
...
+        """Index of the first visible character.
 
-        :return int: the current index
+        :return int: The current index
         """
         return self._current_index
 
     @current_index.setter
-    def current_index(self, new_index):
+    def current_index(self, new_index: int):
```
tekktrik, line 127: If you could add the `None` return type here that would be awesome!

adafruit_display_text/scrolling_label.py
```diff
@@ -110,33 +116,31 @@ def update(self, force=False):
...
 
     @property
-    def full_text(self):
-        """
-        The full text to be shown. If it's longer than `max_characters` then
+    def full_text(self) -> str:
+        """The full text to be shown. If it's longer than ``max_characters`` then
         scrolling will occur as needed.
 
         :return string: The full text of this label.
```
tekktrik, line 138: I know this isn't a part of your PR, but if you wouldn't mind changing this, that would be fantastic!

```suggestion
        :return str: The full text of this label.
```

adafruit_display_text/scrolling_label.py
```diff
@@ -110,33 +116,31 @@ def update(self, force=False):
...
         """
         return self._full_text
 
     @full_text.setter
-    def full_text(self, new_text):
+    def full_text(self, new_text: str):
```
tekktrik, line 143: You can also add the `None` return type here!

docs/api.rst
```diff
@@ -13,3 +13,6 @@
 
 .. automodule:: adafruit_display_text.bitmap_label
    :members:
+
+.. automodule:: adafruit_display_text.scrolling_label
```
tekktrik, line 17: Thank you for catching this!

tekktrik (changes requested):
Thanks for the submission, this is great!  I had some feedback, mostly about the use of `Optional` (and some of these are just thank yous :) ).  Let me know if I can help explain anything!

adafruit_display_text/scrolling_label.py
rgrizzell: It appears as if we'll need to include `FontProtocol` in the [Adafruit_Blinka_Displayio](https://github.com/adafruit/Adafruit_Blinka_Displayio/blob/main/fontio.py) library to get the Sphinx Docs to compile with this change.

rgrizzell: I've added the `FontProtocol` class to the Blinka Displayio library and was able to compile the docs after that. The functions are just specified with a simple `pass` and I need to spend a bit more time to determine what, if anything, needs to be included. I'll create an issue and solicit some feedback over there.

In the meantime, you can test using my fork.

```diff
diff --git a/requirements.txt b/requirements.txt
--- a/requirements.txt	(revision 9417ccc6e033d9a77951835b77bc9f050b210380)
+++ b/requirements.txt	(date 1652758810710)
@@ -3,5 +3,5 @@
 # SPDX-License-Identifier: Unlicense
 
 Adafruit-Blinka
-adafruit-blinka-displayio
+git+https://github.com/rgrizzell/Adafruit_Blinka_Displayio@main#egg=adafruit-blinka-displayio
 adafruit-circuitpython-bitmap-font

```

rgrizzell:
Builds are currently failing until an upstream issue can be resolved.
https://github.com/adafruit/Adafruit_Blinka_Displayio/issues/92

tekktrik (changes requested):
This should be resolved now.  Can you pin the minimum version of `Adafruit_Blinka_displayio` to 0.10.2 in `setup.py` and `requirements.txt` so that anyone upgrading can grab the update library?  Let me know if you'd like help doing so!

rgrizzell:
I think that'll do it. Documentation builds nicely with the latest version of adafruit-blinka-displayio. Thanks for adding the typing.

PR #173, opened 2022-06-15
Use `fontio.FontProtocol` for type annotations

### LINKED ISSUE #172: Use `fontio.FontProtocol` for type annotations
Looking at the library, `fontio.FontProtocol` should should be good to replace `Union[BuiltinFont, BDF, PCF]`!

tekktrik:
Resolves #172 by using `fontio.FontProtocol` for type annotations instead of that huge `Union`.

### DIFF
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -7,6 +7,9 @@
 =======================
 """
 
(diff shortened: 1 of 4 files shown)


tannewt (approved):
Thank you!

Issue #169, opened 2022-05-11
bitmap_label losing direction


TerryCornall:
Hi!
I'm using adafruit-circuitpython-bundle-7.x-mpy-20220507.zip with Circuitpython adafruit-circuitpython-feather_m4_express-en_US-7.2.5.uf2
It's running on an Adafruit feather m4 express with a 2.7" Adafruit Sharp Memory display. (Lovely rig, BTW works very well other than this little issue)

I noticed an issue with adafruit_display_text.bitmap_label just recently where despite the label_direction being set to 'UPR' the label would reset to 'LTR'
After chasing thru my code and making sure I wasn't resetting it to LTR myself, I tried using adafruit_display_text.label rather than a bitmap_label and the problem went away.
Here's a cut-down version (original is waaaaay too big) of the code and it exhibits the same issues.
At first, it appears to work and the label is indeed upwards but if I tell it again that its label_direction is 'UPR' (see down bottom of code marked by ####### )
it suddenly goes LTR orientation and stays that way, despite reporting UPR. It doesn't matter if you do this every cycle or just once, it stays broken.

If I switch to using adafruit_display_text.label rather than a bitmap_label the problem doesn't happen.
Thanks!

```
from adafruit_display_text.bitmap_label import Label #use this instead of adafruit_display_text.label.Label
#from adafruit_display_text.label import Label  #no, use this because bitmap_label loses orientation sometimes
import time
import sharpdisplay
import displayio
import framebufferio
import board
from adafruit_bitmap_font import bitmap_font

class sharp_display():
    def __init__(self):
        self.spi = board.SPI()
        self.disp_cs_pin = board.D4  #D4 is on same side as the SCK, MO, 3,3V, etc so use that
        self.disp_w = 400
        self.disp_h = 240
        self.disp_ncolrs = 2
        self.rotation = 180
        self.framebuffer = sharpdisplay.SharpMemoryFramebuffer(self.spi, self.disp_cs_pin, self.disp_w, self.disp_h)
        self.display = framebufferio.FramebufferDisplay(self.framebuffer, rotation=self.rotation,
                auto_refresh=True)
        self.font = bitmap_font.load_font("fonts/LeagueSpartan-Bold-16.bdf") #big, use scale 1 or 2
        self.palette = displayio.Palette(2)
        self.palette[0] = 0xFFFFFF #so a color of FFFFFF puts a 0 makes display show black
[...]

PR #174, opened 2022-09-10
Fix for bitmap_label losing direction - bug #169

snkYmkrct:
Please see my comment on issue #169

### DIFF
--- a/adafruit_display_text/bitmap_label.py
+++ b/adafruit_display_text/bitmap_label.py
@@ -547,9 +547,12 @@ class Label(LabelBase):
             self._palette.make_transparent(0)
 
     def _set_label_direction(self, new_label_direction: str) -> None:
-        self._prev_label_direction = self._label_direction
-        self._label_direction = new_label_direction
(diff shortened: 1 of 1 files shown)


FoamyGuy (approved):
Looks good to me. 

I think your findings posted in the issue are correct. 

Tested the solution from PR branch successfully with:

```
Adafruit CircuitPython 8.0.0-beta.0 on 2022-08-18; Adafruit PyPortal with samd51j20
Board ID:pyportal
```

In case it may be helpful in the future, here is the test code I used to verify the issue and fix:

```
import time
import board
import terminalio
from adafruit_display_text import bitmap_label

text = "Hello world"
text_area = bitmap_label.Label(terminalio.FONT, text=text)
text_area.x = 10
text_area.y = 10

text_area.label_direction = "UPR"
text_area.label_direction = "UPR"  # second time we set it causes trouble

board.DISPLAY.show(text_area)

val = 0
while True:
    val += 1
    text_area.text = f"Hello World {val}"
    time.sleep(0.3)

```

Issue #175, opened 2022-10-06
Label can take Tuple[int, int, int] for color


tekktrik:
The typing for `Label` (and any parent classes) should be corrected to allow the `color` argument to also take the type `Tuple[int. int ,int]`.  See https://github.com/adafruit/Adafruit_CircuitPython_Simple_Text_Display/pull/13 for reference.

tekktrik:
Resolved by #176

PR #176, opened 2022-10-12
Corrected typing to allow the color arguments to also take a Tuple

isacben:
Hi,

This PR solves Label can take Tuple[int, int, int] for color #175 

Both parameters, `color` and `backgrount_color` can take either an int or a Tuple[int, int, int].

I tested label and bitmap_label with the Adafruit Feather RP2040 and the ST7789 display.

Please let me know if you have any comments!
Thanks!


tekktrik (approved):
Awesome, thanks for following up on this!

isacben:
Thank you! :)

PR #188, opened 2023-03-13
override text property from parent on ScrollingLabel

### LINKED ISSUE #182: Unable to update scrolling_label text outside constructor
Hi,

It seems that you can not update the text of a scrolling_label outside the constructor. 
If outside the constructor but before the showing the display, if I update it to "test" it still retains "Hello World"

eg.

```
import displayio
import terminalio
import time
from adafruit_matrixportal.matrix import Matrix
from adafruit_display_text.scrolling_label import ScrollingLabel

matrix = Matrix()
display = matrix.display


# setup the label
font = terminalio.FONT
splash = displayio.Group()

title_label = ScrollingLabel(
    text="Hello World", font=font, color=0xFFFFFF, scale=1,
    anchored_position=(0, 0), anchor_point=(0, 0),
)

#### Here this does not work
title_label.text = "test"

splash.append(title_label)
display.show(splash)

while True:

   title_label.update()
```

FoamyGuy:
This resolves #182 

As was noted on a comment in the issue `full_text` was the property that was used to change the text rather than `text` like it is with BitmapLabel and Label. So the functionality to change text after creation does exist, but the property name used to do it is not the same as it is on the other Label types.

This was original done that way because I didn't know how to override parent properties while still being able to use the parent implementation of the property in some other parts of code. It turned out that we skirted around that issue by being able to call the internal `_set_text()` function directly rather than using the parents properties.

The change in this PR makes it so that `scrolling_lbl.text` will now function the same as `scrolling_lbl.full_text` which means it will change the text in the scrolling label as likely expected by users who try setting it.

### DIFF
--- a/adafruit_display_text/scrolling_label.py
+++ b/adafruit_display_text/scrolling_label.py
@@ -86,7 +86,7 @@ class ScrollingLabel(bitmap_label.Label):
         if force or self._last_animate_time + self.animate_time <= _now:
 
             if len(self.full_text) <= self.max_characters:
-                self.text = self.full_text
+                super()._set_text(self.full_text, self.scale)
                 self._last_animate_time = _now
                 return
 
@@ -106,8 +106,7 @@ class ScrollingLabel(bitmap_label.Label):
                 _showing_string = "{}{}".format(
                     _showing_string_start, _showing_string_end
                 )
-            self.text = _showing_string
-
+            super()._set_text(_showing_string, self.scale)
             self.current_index += 1
             self._last_animate_time = _now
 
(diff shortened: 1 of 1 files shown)


jposada202020 (approved):
Thank you! :). This work as expected. I used the script in issue #182 given by @VinceVega7. Now you are able to modify the text outside the constructor.

tested on 
```python
Adafruit CircuitPython 8.1.0-beta.0-19-gb6a761335-dirty on 2023-03-10; Adafruit PyPortal Titano with samd51j20
```

PR #191, opened 2023-07-07
bitmaptools.blit refactor support

FoamyGuy:
This adds support for the change this core pr: https://github.com/adafruit/circuitpython/pull/8136

There is still fallback logic so that devices that aren't on the same versions can still both work with the library


tekktrik:
@FoamyGuy is this intended to be drafted still?

FoamyGuy:
@tekktrik thank you. No it's ready now that the core PR is merged.

tekktrik (approved):
Uses a new addition to the API, but in a way that is backward compatible.  Nice!  Looks good to me!

Issue #193, opened 2023-10-25
ValueError from negatives on first character


FoamyGuy:
Reproducer code:
```py
import board
from adafruit_display_text.bitmap_label import Label
from adafruit_bitmap_font import bitmap_font

display = board.DISPLAY
font_file = "fonts/OpenSans-9.bdf"

# Set text, font, and color
text = "Jug"
font = bitmap_font.load_font(font_file)
color = 0xFF00FF

# Create the tet label
text_area = Label(font, text=text, color=color)
text_area.x = 20
text_area.y = 20
display.root_group = text_area

while True:
    pass
```
The font file is from the cleveland art project in the learn guide system. [Download project bundle](https://learn.adafruit.com/elements/3054759/download?type=zip)

The capital J character has a negative x offset in this font file which causes this error to be raised:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "code_bitmap_label_issue.py", line 16, in <module>
  File "adafruit_display_text/bitmap_label.py", line 113, in __init__
  File "adafruit_display_text/bitmap_label.py", line 209, in _reset_text
  File "adafruit_display_text/bitmap_label.py", line 450, in _place_text
  File "adafruit_display_text/bitmap_label.py", line 491, in _blit
ValueError: x must be 0-17
```

It's attempting to use `-1` for `x` due to the negative offset, and being the first character of the string thus starting at x position 0.

I think I have a fix for it that I'll submit in a PR

PR #194, opened 2023-10-26
clamp x value to positive

FoamyGuy:
resolves: #193 

Clamps x value to positive so that it will not attempt to blit out of bounds


makermelissa (approved):
Looks good to me.

PR #195, opened 2023-10-29
Add setuptools to requirements.txt due to Python 3.12

dhalbert:


dhalbert:
There are worse problems. I think we'll need to pin the library builds at 3.11 for now.

PR #196, opened 2023-10-31
chore: allow use of bitmap_label without bitmaptools

### LINKED ISSUE #192: bitmap_label errors if no bitmaptools module present
It appears there's code for handling a fallback, but because `import bitmaptools` is not in a try/except block, it fails.

imnotjames:
in some cases, bitmaptools may not even be used - such as when displayio.Bitmap supports blit.  in others, the module can run without bitmaptools as it has a fallback which works even if it's slower

fixes #192 

tested by ensuring the circuitplayground express - which seems to omit `bitmaptools` - can still blit even if it's slow


FoamyGuy (approved):
Looks good to me. Thank you @imnotjames

PR #208, opened 2024-06-23
Use adafruit_ticks.ticks_ms for scrolling_label animation

RetiredWizard:
As described in adafruit/circuitpython#5411 time.monotonic loses precision over time, which eventually causes the scrolling_label animation to occur every time an update method is called. This can result in the scrolling text suddenly speeding up dramatically.

This PR uses supervisor.ticks_ms and adds a bit of logic to watch for the value rolling over instead of time.monotonic() which should resolve the issue. As mentioned  by Dan in the referenced issue, another option would be to use time.monotonic_ns and similarly watch for the rollover, however apparently not all boards support the monotonic_ns method.

### DIFF
--- a/adafruit_display_text/scrolling_label.py
+++ b/adafruit_display_text/scrolling_label.py
@@ -26,7 +26,7 @@ Implementation Notes
 __version__ = "0.0.0+auto.0"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_Display_Text.git"
 
-from adafruit_ticks import ticks_ms
+import adafruit_ticks
 from adafruit_display_text import bitmap_label
 
 try:
@@ -81,10 +81,10 @@ class ScrollingLabel(bitmap_label.Label):
          Default is False.
         :return: None
         """
-        _now = ticks_ms()
-        if _now < self._last_animate_time:  # ticks_ms has rolled over
-            self._last_animate_time = _now
-        if force or self._last_animate_time + (self.animate_time * 1000) <= _now:
+        _now = adafruit_ticks.ticks_ms()
+        if force or adafruit_ticks.ticks_less(
+            self._last_animate_time + int(self.animate_time * 1000), _now
+        ):
             if len(self.full_text) <= self.max_characters:
                 super()._set_text(self.full_text, self.scale)
                 self._last_animate_time = _now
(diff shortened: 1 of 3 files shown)


RetiredWizard:
The CI didn't like including supervisor. I don't know if that can be fixed in the requirements/toml/??? files somewhere or if it's a bigger issue with the approach.

FoamyGuy:
I think the step that is failing is the docs build. If you add `"supervisor"` to the list here: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/608833a399d21310225c1977c11084927668e732/docs/conf.py#L30 it should be able to complete successfully I think.

RetiredWizard:
> think the step that is failing is the docs build. If you add `"supervisor"` to the list here:

That did the trick :grin: Thanks!

dhalbert (changes requested):
> As mentioned by Dan in the referenced issue, another option would be to use time.monotonic_ns and similarly watch for the rollover, however apparently not all boards support the monotonic_ns method.

A small number of SAMD21 ("M0") boards support `displayio`, but don't support longints and therefore don't support `monotonic_ns`.

One issue is that `supervisor.ticks_ms` is not implemented by Blinka, which would make this not usable if used under Blinka.

I think the solution is to use https://github.com/adafruit/Adafruit_CircuitPython_ticks instead of using `supervisor.ticks_ms()` directly. The `adafruit_ticks library` chooses the ticks mechanism based on what is available.

adafruit_display_text/scrolling_label.py
```diff
@@ -81,8 +81,10 @@ def update(self, force: bool = False) -> None:
...
         """
-        _now = time.monotonic()
-        if force or self._last_animate_time + self.animate_time <= _now:
+        _now = ticks_ms()
+        if _now < self._last_animate_time:  # ticks_ms has rolled over
+            self._last_animate_time = _now
+        if force or self._last_animate_time + (self.animate_time * 1000) <= _now:
             if len(self.full_text) <= self.max_characters:
                 super()._set_text(self.full_text, self.scale)
                 self._last_animate_time = _now
```
dhalbert, line 90: You can use `adafruit_ticks.ticks_diff` or `ticks_less` to handle the wraparound automatically. See https://docs.circuitpython.org/projects/ticks/en/latest/api.html and the library code.

dhalbert (approved):
Looks good, thanks!

Issue #207, opened 2024-05-14
Property `background_tight` not working for `bitmap_label.Label`


SamantazFox:
While trying to figure out [the problem with the width/height properties](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/pull/206), I also noticed that the `background_tight` parameter had no effect on `bitmap_label.Label`.

I haven't had the time to diagnose the issue much further, unfortunately.

![](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/assets/52980486/270a72a0-a579-4af9-85d5-e930d9e111c8)

---

<details><summary>Relevant code</summary>
<p>

```python
import board
import terminalio
import displayio
import adafruit_ili9341
from adafruit_display_text import bitmap_label, label
from adafruit_display_shapes.line import Line

from time import sleep



displayio.release_displays()
display_bus = displayio.FourWire(
	board.SPI(), command=board.D10, reset=board.D11, chip_select=board.D12
)

display = adafruit_ili9341.ILI9341(display_bus, width=240, height=320, rotation=270)


grid = displayio.Group()
for x in range(0, 240, 10): grid.append(Line(x, 0, x, 320, color=0x404040))
for y in range(0, 320, 10): grid.append(Line(0, y, 240, y, color=0x404040))


test_grp = displayio.Group(scale=3)


def add_bb_visualizer(group, txt_area):
	if txt_area._base_alignment:
		y_off = 0
	else:
		y_off = txt_area.height // 2

	x1 = txt_area.x
	y1 = txt_area.y + y_off
	x2 = x1 + txt_area.width
	y2 = y1 - txt_area.height

	# Bounding box
	group.append(Line(x1, y1,  x2, y1, color=0x00FFFF))
	group.append(Line(x2, y1,  x2, y2, color=0xFFFF00))
	group.append(Line(x2, y2,  x1, y2, color=0xFF00FF))
	group.append(Line(x1, y2,  x1, y1, color=0x00FF00))

	# Cross for anchor point
	x0 = txt_area.x
	y0 = txt_area.y
	group.append(Line(x0-2, y0-2,  x0+2, y0+2, color=0xFF0000))
	group.append(Line(x0-2, y0+2,  x0+2, y0-2, color=0xFF0000))


common_kwargs = {
	'text': 'Hello `y',
	'color': 0xFFFFFF,
	'background_color': 0x123456,
	'base_alignment': False,
	# 'background_tight': True,
[...]

PR #209, opened 2024-10-01
Fix for background_tight property not working

snkYmkrct:
Fix for https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/207

When `background_tight` is true, the padding parameters should be ignored.


FoamyGuy (approved):
I tested this fix successfully on a pyportal using a slightly modified version of the reproducer posted in the issue. 

Thank you @snkYmkrct for looking into this and submitting the fix!

Thank you @SamantazFox for the nice reproducer code that made this very easy to test!

PR #210, opened 2024-10-30
Optimize ScrollingLabel update logic to prevent redundant text updates

shubham0x13:
Introduces a check to prevent `_set_text` from being called with unchanged content/text.


FoamyGuy (approved):
Thanks for the improvement @shubham0x13! 

This looks good to me, I did a quick test on a Feather S3 TFT to confirm the new version now does not call _set_text when the text hasn't changed.

PR #211, opened 2024-11-03
Allow empty text and update text instantly

shubham0x13:
- Allowed empty string (`""`) as valid input for the `text` or `full_text` property.
- Forced an immediate update when the `text` or `full_text` property is set, bypassing the `animate_time` delay for instant display updates.


FoamyGuy (approved):
Looks good to me, Thank you!

PR #217, opened 2024-12-03
Optimize LabelBase text setter

shubham0x13:
Added a check to skip redundant `_set_text` calls if the text hasn't changed.


FoamyGuy (approved):
Thank you! This saves a decent little chunk of time when the text is already set to the desired string.

PR #220, opened 2025-05-26
Managing addition of space at end of text

cinderblockgames:
Wrapping max_characters in properties and checking for length of text before adding a space to full_text.

This addresses the situation where the text is shorter or equal to the max_characters, which would add a space when shorter or start scrolling unnecessarily when equal to.  Shorter isn't a big deal when left aligned, but it is noticeable when center or right aligned.


cinderblockgames:
Failed build on `trim trailing whitespace`, but I don't see where I introduced trailing whitespace.  If anyone has time to take a look, I would really appreciate it.

cinderblockgames:
I think I figured it out.  Working on fixing.

PR #223, opened 2025-05-30
Add anchor_label method

shubham0x13:
Added `anchor_label` method which allows easy anchoring/positioning of label based on predefined anchor points and screen size.


shubham0x13:
Thanks! Moved it as suggested and opened a PR there.

PR #237, opened 2026-02-25
add antialias example, move fonts dir into examples.

FoamyGuy:
New example that demonstrates 4 levels of bpp anti-aliasing and the font files that it uses.
image

Also moved the `fonts/` dir into `examples/` instead of the root of the repo which matches the structure from https://github.com/adafruit/Adafruit_CircuitPython_Bitmap_Font/tree/main/examples


tannewt (approved):
Thank you!

PR #239, opened 2026-07-27
fix: avoid orphaning a final wrapped character

### LINKED ISSUE #226: TextBox text length hyphen issue
This was first noticed in: https://github.com/adafruit/Fruit-Jam-OS/issues/62

TextBox seems to automatically add a hyphen at the end of the text if it cannot fit full within the allotted size. But there are cases where there is only a single character remaining in the string and it replaces it with a hyphen. 

Maybe this issue is related to the use of center alignment and/or the last character being an emoji.

giri256:
Fixes #226

When splitting an overlong word, `wrap_text_to_pixels()` always reserved space for a continuation hyphen before testing the next character. On the final character, that could force a character that otherwise fit onto a new line and leave the previous line ending in an unnecessary hyphen.

This reserves hyphen width only when characters remain after the one being tested. True continuations still keep their hyphen, while the final character can use the available line width.

Regression example at width 4:

- before: `['abc-', 'def-', 'g']`
- after: `['abc-', 'defg']`

Verification:

- boundary regression matrix covering exact fit, normal continuation, and the orphaned-final-character case
- `python -m compileall -q adafruit_display_text`
- `ruff format --check adafruit_display_text/__init__.py`
- `ruff check adafruit_display_text/__init__.py`
- clean sdist and wheel build

### DIFF
--- a/adafruit_display_text/__init__.py
+++ b/adafruit_display_text/__init__.py
@@ -119,18 +119,19 @@ def wrap_text_to_pixels(
             cur_part = ""
 
             if wwidth > max_width:
-                for char in word:
+                for char_offset, char in enumerate(word):
                     if newline:
                         extraspace = 0
                         leadchar = ""
                     else:
                         extraspace = swidth
                         leadchar = " "
+                    hyphen_width = measure("-") if char_offset < len(word) - 1 else 0
                     if (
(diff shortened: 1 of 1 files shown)


tannewt (approved):
Thank you!

